Issue #153, opened 2022-06-30
LOD, "cartographic rendering", split screen & full-screen minimap


Indy2222:
Design and implement support for different rendering modes & LOD based on area visible by the/a camera. For example: full 3D -> LOD / simplified 3D -> "cartographic" (mini map style)

Once this is implemented, mini map can be just a fully zoomed out orthographic camera rendered to a texture displayed in the UI.

This would enable some other nice features:

* the ability to have multiple mini maps partially zoomed in on the player's areas of interest
* split screen -> managing in parallel different parts of the map (e.g. a battle at a distant part of the map and the base town)

Issue #135, opened 2022-06-18
Clean-up resources & entiteis after game ends


Indy2222:
Currently, majority of entities and resources inserted during game loading and game play are not removed after the game finishes. This is a memory leak and might lead to incorrect behavior during second game play.

Issue #70, opened 2022-05-04
Implement map selection


Indy2222:
So players can play more than just the one bundled map.

Issue #68, opened 2022-05-04
Make camera parameters configurable


Indy2222:
- Make all or most camera parameters configurable (e.g. movement speed, minimum and maximum zoom, etc.).
- Load & store the configuration to an appropriate path (e.g. XDG_CONFIG).
- Document the configuration format.
- Make sure that issues in the configuration are properly reported (logging and error and fall-backing to default?)

UI for the configuration is not part of the issue. Manual edits to the configuration file are good enough.

This relates to #69.

This is blocked by #243.

PR #223, opened 2022-11-02
higher dezoom + m hotkey to load another map

Vrixyz:
worked on top of [!225](https://github.com/DigitalExtinction/Game/pull/225)

### DIFF
--- a/assets/shaders/terrain.wgsl
+++ b/assets/shaders/terrain.wgsl
@@ -51,7 +51,7 @@ fn draw_circle(
     radius: f32,
 ) -> vec4<f32> {
     let distance: f32 = distance(uv, center);
-    if distance <= (radius + SHAPE_THICKNESS) && radius <= distance  {
+    if distance <= (radius + SHAPE_THICKNESS) && radius <= distance {
         return mix_colors(base, SHAPE_COLOR);
     }
     return base;
@@ -103,7 +103,7 @@ fn nearest(uv: vec2<f32>) -> u32 {
 
         let axis = next.depth % 2u;
         let diff = uv[axis] - node.location[axis];
-        
+
         var close = 2u * next.index + 2u;
         var away = 2u * next.index + 1u;
 
--- a/crates/camera/src/camera.rs
+++ b/crates/camera/src/camera.rs
@@ -23,13 +23,18 @@ const CAMERA_HORIZONTAL_SPEED: InverseSecond = Quantity::new_unchecked(2.0);
 /// Minimum camera distance from terrain achievable with zooming along.
 const MIN_CAMERA_DISTANCE: Metre = Quantity::new_unchecked(20.);
 /// Maximum camera distance from terrain achievable with zooming alone.
-const MAX_CAMERA_DISTANCE: Metre = Quantity::new_unchecked(100.);
+const MAX_CAMERA_DISTANCE: Metre = Quantity::new_unchecked(300.);
+/// Maximum camera distance from terrain achievable with zooming alone.
+const DEFAULT_CAMERA_DISTANCE: Metre = Quantity::new_unchecked(
+    (MIN_CAMERA_DISTANCE.inner() + (MAX_CAMERA_DISTANCE.inner() - MIN_CAMERA_DISTANCE.inner()))
+        * 0.4,
+);
 /// Minimum temporary distance from terrain. Forward/backward camera motion is
 /// smooth within this range. Step adjustment is applied outside of this range.
 const HARD_MIN_CAMERA_DISTANCE: Metre = Quantity::new_unchecked(16.);
 /// Maximum temporary distance from terrain. Forward/backward camera motion is
 /// smooth within this range. Step adjustment is applied outside of this range.
-const HARD_MAX_CAMERA_DISTANCE: Metre = Quantity::new_unchecked(110.);
+const HARD_MAX_CAMERA_DISTANCE: Metre = Quantity::new_unchecked(310.);
 /// Camera moves along forward axis (zooming) at speed `distance *
 /// CAMERA_VERTICAL_SPEED`.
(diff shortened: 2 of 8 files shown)


crates/map/src/io.rs
```diff
@@ -161,17 +161,17 @@ mod test {
...
-
+        let path = Path::new("./de_map_test.tar");
         let tmp_dir = Builder::new().prefix("de_map_").tempdir().unwrap();
         let mut tmp_dir_path = PathBuf::from(tmp_dir.path());
         tmp_dir_path.push("test-map.tar");
 
-        task::block_on(store_map(&map, tmp_dir_path.as_path())).unwrap();
-        let loaded_map = task::block_on(load_map(tmp_dir_path.as_path())).unwrap();
+        task::block_on(store_map(&map, &path)).unwrap();
+        let loaded_map = task::block_on(load_map(&path)).unwrap();
```
Vrixyz, line 170: that was only done to create a fake alternative map ; those changes don't belong there.

crates/loader/src/map.rs
```diff
@@ -48,6 +51,7 @@ fn spawn_map(
     task: Option<ResMut<MapLoadingTask>>,
     mut move_focus_events: EventWriter<MoveFocusEvent>,
     game_config: Res<GameConfig>,
+    entities_to_remove: Query<Entity, Or<(With<Handle<Scene>>, With<RemoveBeforeLoad>)>>,
```
Vrixyz, line 54: we might want a better strategy for removing stuff between loading states, I would err on the side of tagging everything that should not be removed (as I guess there should be a less amount of entities) rather than the opposite, but that was easier to implement.

I might have missed edge cases like the selection rectangle, or projectiles ?

crates/camera/src/camera.rs
```diff
@@ -23,13 +23,13 @@ const CAMERA_HORIZONTAL_SPEED: InverseSecond = Quantity::new_unchecked(2.0);
 /// Minimum camera distance from terrain achievable with zooming along.
 const MIN_CAMERA_DISTANCE: Metre = Quantity::new_unchecked(20.);
 /// Maximum camera distance from terrain achievable with zooming alone.
-const MAX_CAMERA_DISTANCE: Metre = Quantity::new_unchecked(100.);
+const MAX_CAMERA_DISTANCE: Metre = Quantity::new_unchecked(300.);
```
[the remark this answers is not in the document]
Indy2222: You are right, at this point in time this change makes sense.

A bit of a context:

* In the long run, I would like to implement a custom LOD so zooming out a lot is feasible from performance perspective.

  * Until this happens, it might be necessary to change this back to a lower value after we plug in better unit models (with possibly higher triangle counts etc.).

* The appearance / UI has to change on high camera distances: drawing simple colored squares instead of true 3D models, not displaying health bars and selection circles etc. More info here #153.

* Implementing some kind of a mini map (displaying at least all units & buildings & camera) is high on my priority list. That would solve most of the issue you mention (long scroll -> single click on the mini map, getting disoriented -> looking where you are looking at at the mini map).

crates/map/src/io.rs
Indy2222: I've done the same trick. Another PR which extends the `de_tools` crate with some kind of simple "generate map" functionality would be appreciated.

crates/camera/src/camera.rs
```diff
@@ -209,7 +209,7 @@ impl DesiredPoW {
     }
 }
 
-fn setup(mut commands: Commands) {
+fn setup(mut commands: Commands, camera_query: Query<Entity, With<Camera3d>>) {
```
Indy2222, line 212: I will have to thing about it a bit more, but my current thought is that the whole map cycling & game setup has to be done differently:

* Proper clean up after `GameState::Playing` is existed has to be implemented.
* Map selection should be done via main menu. It might be very simplistic at this point (a button for each available map?).
* Generally: once you enter a game (`GameState::Playing`) you cannot change its configuration (like in most RTS games). You have to quit the game first and then start a new game.


My second point is of highest priority IMO.

crates/loader/src/map.rs
Indy2222: I think that the desired solution to this problem is to add on exist system(s) to every plugin so that the responsibility is moved to where the entities / resources were inserted in the first place.

This all thing might not be needed to add map selection (see the other comment): I would just implement simple main menu where you click on a button with a map name to start the game. You would have to restart the game to select another map to decouple proper system clenaups with map selection.

Indy2222 (changes requested):
Thank you for your contribution! It is the first contribution from anyone but me. I really appreciate it.

I see that this is still a draft but I have reviewed it anyways to give you some feedback ASAP.

Indy2222:
Also, please mention this issue in relevant commits: #70

crates/camera/src/camera.rs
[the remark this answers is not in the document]
Indy2222: This is related #68.

assets/shaders/terrain.wgsl
```diff
@@ -51,7 +51,7 @@ fn draw_circle(
     radius: f32,
 ) -> vec4<f32> {
     let distance: f32 = distance(uv, center);
-    if distance <= (radius + SHAPE_THICKNESS) && radius <= distance  {
+    if distance <= (radius + SHAPE_THICKNESS) && radius <= distance {
```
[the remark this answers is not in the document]
Vrixyz: credits goes to my automatic format on save I think :) ; sure I can do that

crates/loader/src/map.rs
Indy2222: Related #135.

crates/camera/src/camera.rs
Indy2222: Related #34.

Indy2222:
Superseded by #259.

Issue #276, opened 2022-11-15
Document Game Object Format


Indy2222:
Currently, object files are placed in `/assets/objects`. Each object is contained in a JSON file with:

- object footprint (a polygon)
- object collider (trimesh)
- altitude (once #262 is done)

Also, document how `de_tools` crate can be used to generate the file (or its parts).

Issue #262, opened 2022-11-13
Control unit altitude properly


Indy2222:
As documented in [CONTRIBUTING.md](https://github.com/DigitalExtinction/Game/blob/main/CONTRIBUTING.md#3d-models), 3D models of units are just above XZ plane. Implement proper control of altitude (which is added to Y coordinate of object translation).

Object altitude should be loaded from object JSON file.

Issue #238, opened 2022-11-06
Design 3D models


Indy2222:
The goal is to come up with first design of 3D models (and their textures and so on). The models should have consistent style with one another and with other aspects of the game.

The list and description of all buildings and units is available in [DESIGN.md](https://github.com/DigitalExtinction/Game/blob/main/DESIGN.md).

PR #278, opened 2022-11-18
Replaced suzanne monkey head with actual tank model.

LarsDu:
Felt bad for this project since it looks like the monkey head is still being used as a placeholder, so here is a hovertank mesh.

Also, the loading code looks a bit overly complex and hard to parse. I'd recommend simplifying things a bit.


Indy2222 (changes requested):
Thank you for the contribution ❤️!

You are right, we are still at the very beginning regarding art and overall visual design. This has to change since it might put people off.

Regarding the simplification, the first step is to document some not-yet-documented parts, e.g. #276. If you have other suggestions, feel free to open a discussion or submit another PR.

The 3D models themselves are already documented here [CONTRIBUTING.md#3d-models](https://github.com/DigitalExtinction/Game/blob/main/CONTRIBUTING.md#3d-models). Though, there is still this small issue which I hope to fix soon #262.

Suggestions:

* [ ] Would it be possible to include `Relates to #238` in the commit message (as per [CONTRIBUTING.md#pull-requests--git](https://github.com/DigitalExtinction/Game/blob/main/CONTRIBUTING.md#pull-requests--git))?
* [ ] Please track the .blend file with LFS (it is almost 1MiB large).
* [ ] Make sure that the bottom of the tank is on XZ plane ([CONTRIBUTING.md#3d-models](https://github.com/DigitalExtinction/Game/blob/main/CONTRIBUTING.md#3d-models)).
* [ ] The muzzle coordinates seems to be off (inside `assets/objects/monk_tank.obj.json`). It should be placed at the tip of the cannon, i.e. arrow 1) in the picture.
* [ ] The convex hull doesn't fully contain the object. See arrow 2) in the picture pointing to the white circumscribed circle around the convex hull.

Note that the issues might have originated in usage of `de_tools` crate (if you used it) due to this bug #279 (just created). Independently to the issue, I suggest to reset model scale (in Blender) to 1:1 before export to GLB (if that is not already the case).

Let me know how you want to proceed. I take over the PR if you want me to.

tank

LarsDu:
It's better if you take over the PR. I'm just here to make a small contribution. Feel free to cut .blend out of source control entirely

PR #280, opened 2022-11-19
Replace suzanne monkey head with actual tank model

Indy2222:
Based on #278 + some modifications.


Indy2222:
Abandoning this: we need a drone model not a tank.

Issue #100, opened 2022-06-02
Add a Minimap


Indy2222:
- It displays all buildings and units as simple colored shapes (e.g. rectangles).
- ~It is rotated with the camera.~ -> #374

Issue #22, opened 2022-04-03
Implement Unit Manufacturing


Indy2222:
- When a single factory building is selected, a menu with available units is shown. When the user clicks on a unit, it is added to a manufacturing queue.
- For each factory, there is a queue of units to be manufactured. The queue is gradually processed.

PR #317, opened 2023-01-14
feat(controller): implement basic HUD

Polostor:
Added basic HUD with possible extensions in both top and bottom designs
image


crates/controller/src/mouse.rs
```diff
@@ -5,6 +5,7 @@ use bevy::{
 };
 use de_core::{screengeom::ScreenRect, stages::GameStage, state::GameState};
 use iyes_loopless::prelude::*;
+use de_hud::HUDInteraction;
```
Indy2222, line 8: I would change this to `HudInteraction` so only initial letter of the abbreviation is capitalized.

crates/hud/src/draft.rs
```diff
@@ -0,0 +1,19 @@
...
+    let point = Vec3::new(0.0, 0.0, 0.0);
+    info!("OnButtonClicked: Spawning object {}", building_type);
+    commands.spawn(DraftBundle::new(
+        building_type,
+        Transform {
+            translation: point,
+            ..Default::default()
+        },
+    ));
+}
```
Indy2222, line 19: Please add an end-of-line here.

crates/hud/src/universal_ui.rs
```diff
@@ -0,0 +1,91 @@
...
+        size: Size {
+            width: Val::Percent(100.0),
+            height: Val::Percent(100.0),
+        },
+        align_items: AlignItems::Center,
+        flex_direction: FlexDirection::Column,
+        justify_content: JustifyContent::SpaceEvenly,
+        ..Style::default()
+    }
+}
```
Indy2222, line 91: Ditto, please make sure that all files end with a new line.

crates/hud/src/interaction.rs
```diff
@@ -0,0 +1,39 @@
...
+    BuildBuilding(BuildingType)
+}
+
+/// Handles what should when then a menu item in the main menu is selected
+/// by the user.
+///
+/// # Arguments
+///
+/// * `query` - The query for the buttons in the menu.
+pub fn handle_item_click(
```
Indy2222, line 19: I generally try to place all systems alongside the plugin which adds them. Note that `hud` crate implements a PluginGroups which can add multiple plugins. So I suggest to create a new plugin in each module so they are not all added from `plugin.rs`.

Together with that, `plugin.rs` should be renamed to something more specific since it would no longer be the only plugin in the crate.

crates/controller/src/mouse.rs
```diff
@@ -22,7 +23,9 @@ impl Plugin for MousePlugin {
                     .with_system(
                         update_position
                             .run_in_state(GameState::Playing)
-                            .label(MouseLabels::Position),
+                            .label(MouseLabels::Position)
+                            .after(HUDInteraction::Click)
```
Indy2222, line 27: Why is this needed? Is it because you rely on mouse input clearing (`mouse.clear();`) inside the system with this label?

I think that this is easy to miss subtlety (a mine which will cause bugs in the future). I think that we should come up with a more robust input handling mechanism.

Also, what happens if the mouse is clicked while over HUD but no button?

crates/hud/src/draft.rs
```diff
@@ -0,0 +1,19 @@
+use bevy::prelude::{Commands, info, Transform, Vec3};
+
+use de_core::objects::BuildingType;
+use de_spawner::DraftBundle;
+
+pub fn place_draft(
```
Indy2222, line 6: Pretty much the same thing is already implemented inside controller (when P or B is pressed). I suggest to create an event for this inside controller and send that event instead (from both the keyboard handling system in controller and the button handling system in HUD).

This deserves a bit more thought to avoid a circular dependency (between hud and controller). Maybe the two should be merged together? What do you think?

crates/hud/src/hud.rs
```diff
@@ -0,0 +1,124 @@
...
+use de_core::objects::BuildingType;
+
+use crate::interaction::BuildBuildingButton::BuildBuilding;
+use crate::universal_ui::{create_button_bundle, create_detail_node_bundle, create_side_node_bundle, create_text_in_button_bundle, universal_style_full};
+
+pub fn spawn_hud(
+    mut commands: Commands,
+    asset_server: ResMut<AssetServer>,
+) {
+    let font: Handle<Font> = asset_server.load("fonts/Fira_Mono/FiraMono-Medium.ttf");
```
Indy2222, line 15: I hoped to have all the low-level / building-block UI functionality inside the UI crate and then use it here to build up the HUD.

crates/hud/src/universal_ui.rs
```diff
@@ -0,0 +1,91 @@
+use bevy::asset::Handle;
+use bevy::prelude::{ButtonBundle, Color, NodeBundle, TextBundle};
+use bevy::text::{Font, HorizontalAlign, Text, TextAlignment, TextStyle, VerticalAlign};
+use bevy::ui::{AlignItems, AlignSelf, BackgroundColor, FlexDirection, JustifyContent, Size, Style, UiRect, Val};
+use bevy::utils::default;
+
+pub fn create_side_node_bundle(color: Color) -> NodeBundle {
```
Indy2222, line 7: Ditto, please have a look at ui crate which is designed to implement all the low level UI building blocks.

Indy2222 (changes requested):
Please mention tickets #22 and #100 in commit description.

crates/hud/src/draft.rs
Polostor: I do agree, I have created new [PR](https://github.com/DigitalExtinction/Game/pull/334) for that

crates/controller/src/mouse.rs
Polostor: My idea was almost what you described. I wanted to create the set of steps by which the click would be processed. But as you correctly pointed out it would not cover the clicks on all visible HUD. Will need to rethink this..

Polostor:
**This PR is still WIP**.. I'll still need to rework some of the stuff but I wanted to have it on latest before I do so..

crates/controller/src/hud.rs
```diff
@@ -79,3 +102,32 @@ fn process_events(
...
+                flex_direction: FlexDirection::Row,
+                align_items: AlignItems::FlexEnd,
+                position_type: PositionType::Absolute,
+                size: Size {
+                    width: Val::Percent(100.0),
+                    height: Val::Percent(100.0),
+                },
+                ..Style::default()
+            },
+            background_color: BackgroundColor::from(Color::NONE),
```
Indy2222, line 124: `Color::NONE.into()` would be more concise & idiomatic in this context (the type is derived from `NodeBundle` struct).

crates/controller/src/hud.rs
```diff
@@ -79,3 +102,32 @@ fn process_events(
...
+            style: Style {
+                justify_content: JustifyContent::SpaceBetween,
+                flex_direction: FlexDirection::Row,
+                align_items: AlignItems::FlexEnd,
+                position_type: PositionType::Absolute,
+                size: Size {
+                    width: Val::Percent(100.0),
+                    height: Val::Percent(100.0),
+                },
+                ..Style::default()
```
Indy2222, line 122: `Style::default()` -> `default()` is the idiomatic way to do it.

crates/controller/src/hud_interaction.rs
```diff
@@ -0,0 +1,32 @@
...
+};
+
+#[derive(Component, Clone, Copy)]
+pub(crate) enum HudButtonAction {
+    Build(BuildingType),
+}
+
+pub(crate) fn hud_button_system(
+    interactions: Query<(&Interaction, &HudButtonAction), Changed<Interaction>>,
+    counter: Res<ObjectCounter>,
```
Indy2222, line 18: Ditto, with this modification, it would make sense to move the counter logic into the receiver of the draft event.

crates/controller/src/hud_interaction.rs
```diff
@@ -0,0 +1,32 @@
...
+    pointer: Res<Pointer>,
```
Indy2222, line 19: It is clear that the pointer is over HUD -> I would change initial placement to the middle of the screen. Note that this might require modification of the camera crate to somehow expose focus terrain point (it is already there but not public and with nice API).

Indy2222 (changes requested):
* Can you split this into 2 PRs -- first include the empty HUD, second implement draft placement logic / buttons.
* Please mark the non-transparent nodes (but not their descendants) with a marker trait & use this inside `mouse.rs` (`update_position()` system) to set mouse position to None (similar to when cursor leaves the window).

Indy2222:
This PR has been superseded by #365 and follow ups.

PR #318, opened 2023-01-15
Double click to select all entities of the same type

### LINKED ISSUE #234: Select all visible entities of the same type
Double click on an entity should select all visible entities of the same type.

0HyperCube:
Closes #234

I also made it only select entities on the screen, similar to StarCraft2.


crates/controller/src/command.rs
```diff
@@ -162,6 +170,31 @@ fn left_click_handler(
         };
 
         let event = match pointer.entity().filter(|&e| playable.contains(e)) {
+            Some(entity) if is_double_click => {
+                let selected_entity_type = playable.get_component::<ObjectType>(entity).unwrap();
```
Indy2222, line 174: There is already a very similar functionality in [`select_in_area()`](https://github.com/DigitalExtinction/Game/blob/main/crates/controller/src/areaselect.rs#L58). So I suggest to generalize that system so it can filter based on `Option<ObjectType>` (new part of `SelectInRectEvent`) and thus this system would only send `SelectInRectEvent` with a rectangle consisting of full screen.

Indy2222 (changes requested):
Thank you for the work! I have left a few comments.

Please document the new functionality in `TUTORIAL.md`.

crates/controller/src/command.rs
0HyperCube: Thanks, I didn't see that function.

crates/controller/src/areaselect.rs
```diff
@@ -66,6 +79,10 @@ fn select_in_area(
         let event_frustum = screen_frustum.rect(in_event.rect());
         let entities: Vec<Entity> = candidates
             .iter()
+            .filter(|(_, &object_type, _)| {
+                in_event.filter_object_type().is_none()
```
Indy2222, line 83: I would slightly prefer something like this: `in_event.filter_object_type().map_or(true, |filter| filter == object_type)`

crates/controller/src/mouse.rs
```diff
@@ -218,6 +237,13 @@ fn update_buttons(
                     match drag_resolution {
                         DragResolution::Point(_) => {
                             clicks.send(MouseClicked::new(event.button));
```
Indy2222, line 239: I would send this event only if double click was not triggered so the two events are never send at once. I have no strong opinion though.

Indy2222 (changes requested):
Two minor optional comments + please squash all three commit to a single one. Otherwise SGTM. Thank you!!

crates/controller/src/mouse.rs
0HyperCube: In js, a second click event is sent followed by a double click event.

Indy2222: I didn't know that. Better to stick to conventions so lets keep it as is.

Indy2222 (approved):
SGTM

crates/controller/src/mouse.rs
```diff
@@ -209,7 +225,10 @@ fn update_buttons(
     mut mouse_state: ResMut<MouseDragStates>,
     mut input_events: EventReader<MouseButtonInput>,
     mut clicks: EventWriter<MouseClicked>,
+    mut double_clicks: EventWriter<MouseDoubleClicked>,
```
Indy2222, line 228: Clippy is failing on this function because it takes too many arguments. I just realized that we can keep sending only click events from the system and create another one which listens on the events and creates a DoubleClick if two or more come within a short timespan.

Indy2222:
Note that this #334 was merged in the meantime. It would be safer to rebase the branch on newest `main` and re-test.

Indy2222 (approved):

crates/controller/src/command.rs
```diff
@@ -146,6 +163,34 @@ fn right_click_handler(
...
+    pointer: Res<Pointer>,
+    playable: Query<&ObjectType, With<Playable>>,
+    drafts: Query<(), With<Draft>>,
+    mut select_in_rect_events: EventWriter<SelectInRectEvent>,
+) {
+    if !drafts.is_empty() {
+        return;
+    }
+    let selection_mode = if keys.pressed(KeyCode::LControl) {
+        SelectionMode::AddToggle
```
Indy2222, line 177: I would use `Add` instead of `AddToggle` (the latter makes sense for single click but not double click).

Indy2222 (changes requested):
I've been testing the changes. Everything works fine but I discovered one small usability issue (`Add` vs `AddToggle` -- see my code comment).

I am sorry that I haven't realized this sooner!

Indy2222 (approved):

PR #339, opened 2023-01-18
Fix entity selection

Indy2222:


Indy2222:
FYI @0HyperCube there was a bug in the selection code (not introduced by you) which surfaced after the changes you made.

PR #340, opened 2023-01-18
Improve double click detection

Indy2222:


0HyperCube:
Forgot double clicking needed to be at the same position ;)

Indy2222:
> Forgot double clicking needed to be at the same position ;)

No worries.

PR #343, opened 2023-01-19
Change draft colour based on allowed state

### LINKED ISSUE #142: Change color (e.g. red / green) of drafts based on possibility to build them on the spot
This is a follow up after #23.

Users need to immediately see whether a building can be build on a spot. Change color of the buildings to semi-transparent red/green based on collisions with other objects, out-of-map constraint, and so on.

0HyperCube:
Closes #142

### DIFF
--- a/crates/spawner/src/draft.rs
+++ b/crates/spawner/src/draft.rs
@@ -5,6 +5,7 @@
 //! visualized by the plugin.
 
 use bevy::prelude::*;
+use bevy::scene::SceneInstance;
 use de_core::{
     objects::{ActiveObjectType, BuildingType, MovableSolid, ObjectType, StaticSolid},
     projection::ToFlat,
@@ -24,6 +25,9 @@ use parry3d::math::Isometry;
 const MAP_PADDING: f32 = 2. * EXCLUSION_OFFSET + 0.1;
 const MAP_OFFSET: Vector<f32> = Vector::new(MAP_PADDING, MAP_PADDING);
 
+const VALID_PLACEMENT: Color = Color::rgba(0.2, 0.8, 0.2, 0.7);
+const INVALID_PLACEMENT: Color = Color::rgba(0.86, 0.08, 0.24, 0.7);
+
 pub(crate) struct DraftPlugin;
 
 impl Plugin for DraftPlugin {
@@ -37,7 +41,20 @@ impl Plugin for DraftPlugin {
             update_draft
                 .run_in_state(GameState::Playing)
                 .after(IndexLabel::Index),
-        );
+        )
+        .add_system_to_stage(
+            GameStage::PostUpdate,
+            check_draft_loaded
+                .run_in_state(GameState::Playing)
+                .after(IndexLabel::Index),
+        )
+        .add_system_to_stage(
+            GameStage::PostUpdate,
+            update_draft_colour
+                .run_in_state(GameState::Playing)
+                .after(IndexLabel::Index),
+        )
+        .add_startup_system(insert_materials);
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/spawner/src/draft.rs
```diff
@@ -126,3 +142,69 @@ fn update_draft(
...
+#[derive(Clone, Resource)]
+struct DraftMaterials {
+    valid_placement: Handle<StandardMaterial>,
+    invalid_placement: Handle<StandardMaterial>,
+}
+
+fn insert_materials(mut commands: Commands, mut materials: ResMut<Assets<StandardMaterial>>) {
+    commands.insert_resource(DraftMaterials {
+        valid_placement: materials.add(Color::rgba(0.2, 0.8, 0.2, 0.7).into()),
+        invalid_placement: materials.add(Color::rgba(0.86, 0.08, 0.24, 0.7).into()),
```
0HyperCube, line 156: Should these colours be declared as constants somewhere?

crates/spawner/src/draft.rs
```diff
@@ -126,3 +142,69 @@ fn update_draft(
...
+    mut standard_materials: Query<&mut Handle<StandardMaterial>>,
+    scene_spawner: Res<SceneSpawner>,
+    draft_materials: Res<DraftMaterials>,
+) {
+    for (draft, children, mut draft_highlight) in &mut draft_query {
+        let allowed = draft.allowed();
+        // Skip assigning material if the allowed state hasn't changed
+        if draft_highlight.previously_allowed == Some(allowed) {
+            continue;
+        }
```
0HyperCube, line 191: I'm struggling to get this working with change detection as for the first few frames after the draft is spawned, the scene doesn't exist.

crates/spawner/src/draft.rs
Indy2222: Yes, I would create constants at the top of the file.

crates/spawner/src/draft.rs
```diff
@@ -126,3 +142,69 @@ fn update_draft(
...
+        invalid_placement: materials.add(Color::rgba(0.86, 0.08, 0.24, 0.7).into()),
+    });
+}
+
+// Assign the appropriate allowed to all entities in the spawned glb scene
+fn update_object_material(
+    allowed: bool,
+    entities: impl Iterator<Item = Entity>,
+    standard_materials: &mut Query<&mut Handle<StandardMaterial>>,
+    draft_materials: &Res<DraftMaterials>,
```
Indy2222, line 165: `&Res<DraftMaterials>` -> `&DraftMaterials` would be nicer (you can get it with `res.as_ref()`.

crates/spawner/src/draft.rs
Indy2222: I see. I would create a ticket for scene pre-loading (e.g. during loading stage) and reference the ticket here in a comment. Using change detection would be nicer.

Indy2222 (changes requested):

crates/spawner/src/draft.rs
0HyperCube: I managed to get the change detection working by adding a system that checks with a `SceneInstance` component is added.

crates/spawner/src/draft.rs
```diff
@@ -126,3 +143,71 @@ fn update_draft(
...
+
+// Assign the appropriate allowed to all entities in the spawned glb scene
+fn update_object_material(
+    allowed: bool,
+    entities: impl Iterator<Item = Entity>,
+    standard_materials: &mut Query<&mut Handle<StandardMaterial>>,
+    draft_materials: &DraftMaterials,
+) {
+    for entity in entities {
+        let Ok(mut material_handle) = standard_materials.get_mut(entity) else{
```
Indy2222, line 169: There is a forgotten space after `else`.

Indy2222 (changes requested):
Could you change commit description (not the first line) so it includes `Fixes #142` to properly reference the issue (so it is automatically closed).

crates/spawner/src/draft.rs
Indy2222: Cool, I like the current solution!

Indy2222 (approved):

PR #353, opened 2023-01-27
Lobby: use map hash for map identification

Indy2222:


Polostor:
Fixed #354 🥳

Indy2222:
> Fixed #354

How? It does not look related to me.

PR #365, opened 2023-01-29
feat(controller): add simple blocking HUD

Polostor:
Added initial HUD without actions #22
Added node for minimap #100

### DIFF
--- a/crates/controller/src/hud.rs
+++ b/crates/controller/src/hud.rs
@@ -1,6 +1,13 @@
-use bevy::prelude::*;
-use de_core::{screengeom::ScreenRect, stages::GameStage, state::AppState};
+use bevy::{ecs::system::SystemParam, prelude::*};
+use de_core::{
+    screengeom::ScreenRect,
+    stages::GameStage,
+    state::{AppState, GameState},
+};
+use glam::Vec3Swizzles;
 use iyes_loopless::prelude::*;
+
+use crate::hud_components;
 
 const SELECTION_BOX_COLOR: Color = Color::rgba(0., 0.5, 0.8, 0.2);
 
@@ -9,6 +16,7 @@ pub(crate) struct HudPlugin;
 impl Plugin for HudPlugin {
     fn build(&self, app: &mut App) {
         app.add_event::<UpdateSelectionBoxEvent>()
+            .add_enter_system(GameState::Playing, spawn_hud)
             .add_system_set_to_stage(
                 GameStage::PostUpdate,
                 SystemSet::new().with_system(process_events.run_in_state(AppState::InGame)),
@@ -77,5 +85,40 @@ fn process_events(
                 }
             }
         }
+    }
+}
+
+pub fn spawn_hud(mut commands: Commands) {
+    hud_components::spawn_details(&mut commands);
+    hud_components::spawn_action_bar(&mut commands);
+    hud_components::spawn_map(&mut commands);
+}
+
+/// Top-level non-transparent UI node. All such nodes are marked with this component and no descendants have it attached
+#[derive(Component)]
+pub struct HudTopVisibleNode;
+
+#[derive(SystemParam)]
+pub(crate) struct HudNodes<'w, 's> {
+    hud: Query<'w, 's, (&'static GlobalTransform, &'static Node), With<HudTopVisibleNode>>,
+    windows: Res<'w, Windows>,
+}
+
+impl<'w, 's> HudNodes<'w, 's> {
+    pub(crate) fn contains_point(&mut self, point: &Vec2) -> bool {
+        let window = self.windows.get_primary().unwrap();
+        let win_size = Vec2::new(window.width(), window.height());
+        self.hud.iter().any(|(box_transform, node)| {
+            // WARNING: This is because mouse y starts on bottom, GlobalTransform on top
+            let mouse_position = Vec2::new(point.x, win_size.y - point.y);
+
+            let box_size = node.size();
+            let box_transform: Vec3 = box_transform.translation();
+            // WARNING: This is because GlobalTransform is centered, width/2 to left and to right, same on vertical
+            let box_position = box_transform.xy() - box_size / 2.;
+
+            mouse_position.cmpge(box_position).all()
+                && mouse_position.cmple(box_position + box_size).all()
+        })
     }
 }
--- a/crates/controller/src/hud_components.rs
+++ b/crates/controller/src/hud_components.rs
@@ -1,0 +1,77 @@
+use bevy::prelude::*;
+
+use crate::hud::HudTopVisibleNode;
+
+const HUD_COLOR: Color = Color::BLACK;
+
+pub(crate) fn spawn_details(commands: &mut Commands) {
+    commands.spawn((
+        NodeBundle {
+            style: Style {
+                size: Size {
+                    width: Val::Percent(20.),
+                    height: Val::Percent(30.),
+                },
+                position_type: PositionType::Absolute,
+                position: UiRect::new(
+                    Val::Percent(0.),
+                    Val::Percent(20.),
+                    Val::Percent(70.),
+                    Val::Percent(100.),
+                ),
+                ..default()
+            },
+            background_color: HUD_COLOR.into(),
+            ..default()
+        },
+        HudTopVisibleNode,
+    ));
+}
+
+pub(crate) fn spawn_action_bar(commands: &mut Commands) {
+    commands.spawn((
+        NodeBundle {
+            style: Style {
+                size: Size {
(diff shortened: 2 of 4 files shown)


crates/controller/src/hud.rs
```diff
@@ -79,3 +89,30 @@ fn process_events(
...
+    commands.spawn(
+        NodeBundle {
+            style: Style {
+                justify_content: JustifyContent::SpaceBetween,
+                flex_direction: FlexDirection::Row,
+                align_items: AlignItems::FlexEnd,
+                position_type: PositionType::Absolute,
+                size: Size {
+                    width: Val::Percent(100.0),
+                    height: Val::Percent(100.0),
```
Indy2222, line 105: Cannot the height be set to the height of the tallest component?

crates/controller/src/hud_components.rs
```diff
@@ -0,0 +1,52 @@
...
+    commands: &mut ChildBuilder,
+) {
+    commands.spawn((side_node_bundle(HUD_COLOR), HudVisibleNode));
+}
+
+fn side_node_bundle(color: Color) -> NodeBundle {
+    NodeBundle {
+        style: Style {
+            size: Size { width: Val::Percent(20.), height: Val::Percent(30.) },
+            align_items: AlignItems::Stretch,
```
Indy2222, line 37: Why are non-default values needed? At a first glance, it seems like only `size` is relevant in current state.

crates/controller/src/mouse.rs
```diff
@@ -212,16 +212,47 @@ enum DragResolution {
...
+    hud: Query<(&GlobalTransform, &Node), With<HudVisibleNode>>,
+    mut mouse: ResMut<MousePosition>,
+) {
     let window = windows.get_primary().unwrap();
     mouse.set_position(
         window
             .cursor_position()
+            .and_then(|position| {
+                let win_size = Vec2::new(window.width(), window.height());
+                let found = hud.iter().find(|(transform, node)| {
```
Indy2222, line 226: You can replace `.find()` with `.any()` which is what you want here.

crates/controller/src/mouse.rs
```diff
@@ -212,16 +212,47 @@ enum DragResolution {
...
+                if found.is_none() { Some(position) } else { None }
+            })
             .map(|position| position / Vec2::new(window.width(), window.height()))
             .map(|normalised_position| normalised_position.clamp(Vec2::ZERO, Vec2::ONE)),
     );
 }
 
+fn is_in(
+    box_transform: &GlobalTransform,
+    box_size: &Node,
```
Indy2222, line 238: `box_size` -> `node` would be a more appropriate argument name.

crates/controller/src/mouse.rs
```diff
@@ -212,16 +212,47 @@ enum DragResolution {
...
+    win_size: Vec2,
+    mouse_position: Vec2,
+) -> bool {
+    // WARNING: This is because mouse y starts on bottom, GlobalTransform on top
+    let mouse_position = Vec2::new(mouse_position.x, win_size.y - mouse_position.y);
+
+    let box_size = box_size.size();
+    let box_transform: Vec3 = box_transform.translation();
+    // WARNING: This is because GlobalTransform is centered, width/2 to left and to right, same on vertical
+    let box_position = Vec2::new(box_transform.x, box_transform.y) - box_size / Vec2::splat(2.);
```
Indy2222, line 248: * `Vec2::new(box_transform.x, box_transform.y)` -> `box_transform.xy()` should work.
* Isn't `box_size / Vec2::splat(2.)` the same as `box_size / 2.`?

crates/controller/src/mouse.rs
```diff
@@ -212,16 +212,47 @@ enum DragResolution {
...
+
+    box_position.x <= mouse_position.x
```
Indy2222, line 250: `mouse_position.cmpge(box_position) && mouse_position.cmple(box_position + box_size)` is more concise.

crates/controller/src/mouse.rs
```diff
@@ -212,16 +212,47 @@ enum DragResolution {
...
             .map(|position| position / Vec2::new(window.width(), window.height()))
             .map(|normalised_position| normalised_position.clamp(Vec2::ZERO, Vec2::ONE)),
     );
 }
 
+fn is_in(
+    box_transform: &GlobalTransform,
+    box_size: &Node,
+    win_size: Vec2,
+    mouse_position: Vec2,
```
Indy2222, line 240: This function does not care whether it is position of mouse or something else, so I would rename the argument to `point` or something generic.

crates/controller/src/mouse.rs
```diff
@@ -212,16 +212,47 @@ enum DragResolution {
...
+                    is_in(transform, node, win_size, position)
+                });
+                if found.is_none() { Some(position) } else { None }
+            })
             .map(|position| position / Vec2::new(window.width(), window.height()))
             .map(|normalised_position| normalised_position.clamp(Vec2::ZERO, Vec2::ONE)),
     );
 }
 
+fn is_in(
```
Indy2222, line 236: `is_in` -> `contains_point()` would IMHO be a better name.

crates/controller/src/mouse.rs
```diff
@@ -212,16 +212,47 @@ enum DragResolution {
     Rect(ScreenRect),
 }
 
-fn update_position(windows: Res<Windows>, mut mouse: ResMut<MousePosition>) {
+fn update_position(
+    windows: Res<Windows>,
+    hud: Query<(&GlobalTransform, &Node), With<HudVisibleNode>>,
```
Indy2222, line 217: I would experiment with making this a `SystemParam` to make the API nicer. The usage would look like this:

```rust
fn update_position(
    windows: Res<Windows>,
    hud: HudNodes,
    mut mouse: ResMut<MousePosition>,
) {
...
            .filter(|position| hud.contains_point(position))
}
```

crates/controller/src/mouse.rs
```diff
@@ -212,16 +212,47 @@ enum DragResolution {
...
+    mut mouse: ResMut<MousePosition>,
+) {
     let window = windows.get_primary().unwrap();
     mouse.set_position(
         window
             .cursor_position()
+            .and_then(|position| {
```
Indy2222, line 224: `.and_then()` is not needed, `.filter()` would suffice.

crates/controller/src/hud.rs
```diff
@@ -79,3 +89,30 @@ fn process_events(
         }
     }
 }
+
+pub fn spawn_hud(
+    mut commands: Commands,
+) {
+    commands.spawn(
```
Indy2222, line 96: Is this needed? I think that setting absolute position to all three children and not using the parent would be more concise.

crates/controller/src/mouse.rs
```diff
@@ -288,3 +319,6 @@ fn check_double_click(
         *last_click_position = Some(mouse_clicked.position());
     }
 }
+
+#[derive(Component)]
+pub(crate) struct HudVisibleNode;
```
Indy2222, line 324: What about `HudTopVisibleNode`? Also, please document what this component means.

And I would place it alongside HUD as it is where it logically belongs.

Indy2222 (changes requested):
I had a lot of comments, but the overall idea looks good!

crates/controller/src/mouse.rs
Polostor: It had to be using other method ( `some.cmpge(other).all()` ) but great advice!

crates/controller/src/mouse.rs
Polostor: It does belong to HUD, but as you written somewhere before you try to have all systems used in one plugin in single file.. which makes me confused as to where to put this `HudTopVisibleNode` and the newly created `SystemParam` called `HudNodes`..

Polostor: And please where to document this? Is it sufficient in **code as a comment** eg. `\\\ Any HUD element that is the parent Node and should is not supposed to be draggable and other  interaction with in-game objects` or somewhere more detailed comment in documentation files like `CONTRIBUTING.md`?

Indy2222: Code comment is sufficient. `Top-level non-transparent UI node. All such nodes are marked with this component and no descendants have it attached` or something like that.

Indy2222: I would move this and the system parameter to the different module. The component does not have to be used from mouse whatsoever. It has to be temporarily public due to the limitations of system param derive, but that will change with Bevy v10.

Indy2222 (dismissed):
SGTM. I will test it later today.

Indy2222 (approved):

PR #368, opened 2023-01-31
feat(controller/hud): add initial in-game map

Polostor:


crates/controller/src/hud/map.rs
```diff
@@ -0,0 +1,30 @@
+use bevy::prelude::{default, AlignSelf, Color, Component, NodeBundle, Size, Style, Val};
+
+pub(crate) struct MapBundle {
```
Indy2222, line 3: * It is not a bundle, since you don't directly spawn it with Bevy.
* Having a struct like this is overkill, it can be a function.
* Further: I would construct & spawn the node directly in `        .with_children(|parent| {`

crates/controller/src/hud/map.rs
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    pub(crate) fn build(&self) -> NodeBundle {
+        NodeBundle {
+            style: Style {
+                size: Size {
+                    width: Val::Percent(100.),
+                    height: Val::Percent(100.),
+                },
+                align_self: AlignSelf::Stretch,
```
Indy2222, line 19: Alternatively, do not pass `size` and `align_self` but instead set `position: UiRect::all(Val::Percent(0.))`, something like that should work.

crates/controller/src/hud/map.rs
```diff
@@ -0,0 +1,30 @@
...
+            },
+            background_color: self.background.into(),
+            ..default()
+        }
+    }
+}
+
+/// Node marker for a parent drawing map details.
+#[derive(Component)]
+pub(crate) struct MapNode;
```
Indy2222, line 30: This does not serve any purpose at the moment. I would delete it.

Indy2222 (changes requested):
Nice. I've left a few notes.

Now the hard part: how to draw into it?

crates/controller/src/hud/map.rs
Polostor: Making it a function seems reasonable, but as of my vision... I thought about it as an **object made of two parts**.. **First**, there is (possibly) **a static part**, that should be drawn, like terrain, at the moment this part is drawn as a `color`. The **second** part should be **dynamic** and it should be made of small dots or whatever the design, and moved around the same as an object.. When _the object_ is moving, this same action should trigger system that is projected on a mini-map  to move a dot around as well..

crates/controller/src/hud/map.rs
Polostor: Based on `mapselection` in a `menu` crate there is at least one more parameter needed. Having either `align_self` or as advised `position` works..

crates/controller/src/hud/map.rs
Polostor: I know, I know.. YAGNI.. But this is _Marker_ Trait which should be there for adding dynamic children, based on additional _Marker_ for `object`, like `MinimapDrawableObject`, to be drawn and moved on the minimap..

crates/controller/src/hud/map.rs
Indy2222: It can be easily changed once there is something drawn to the rect. We need to research how to do it and the final implemented design will almost certainly differ from what you have described (so this code won't survive). Thus I would stick to the YAGNI principle and won't do it this way.

crates/controller/src/hud/map.rs
Indy2222: Once needed, you can add it.

Indy2222 (approved):

PR #370, opened 2023-02-01
Added dron model

jirikrek:


jirikrek:
#238
https://user-images.githubusercontent.com/52002739/216167680-d2be5738-ca9b-4aeb-b9ab-fa6b53756399.mp4

Indy2222:
Thank you for the model.

* How many triangles / polygons does the model have? Note that we need to render hundreds of them in real-time.
* Can you make it smaller, ideally it should fit into a ball with 1m radius.
* Would it be possible to move the laser cannon to the bottom? This way, it won't be able to shoot downwards from the flight.
* Can you make sure that the drone is properly placed on the ground plane? I.e. it should not be crossed by ground plane.
* We can use this model for the base attacker drone. Thus I would replace the current `attacker.glb` file.
* We need to modify the [object file](https://github.com/DigitalExtinction/Game/tree/main/assets/objects) as well (the collider and footprint of the drone).
* For more info on any topic, please refer to [docs.de-game.org](https://docs.de-game.org/overview/) and [`CONTRIBUTING.md`](https://github.com/DigitalExtinction/Game/blob/main/CONTRIBUTING.md).

Issue #90, opened 2022-05-31
Logging to file


Indy2222:
Log to a file. To output could be used for example for bug reports.

PR #444, opened 2023-05-04
Add logging to file functionality

JackCrumpLeys:
Log to a /logs directory and compress old logs. #90


JackCrumpLeys:
I think some more discussion may be needed around log filtering and weather we want a really verbose log for crashes. I have tried to make it easy to just subscribe new writers onto the tracing log. It is a little bit against the overall layout of the code-base to put so much code into the main.rs, do we need to move it?

Cargo.toml
```diff
@@ -46,9 +46,14 @@ de_pathing.workspace = true
 de_signs.workspace = true
 de_spawner.workspace = true
 de_terrain.workspace = true
+de_log.workspace = true
```
Indy2222, line 49: Please keep all blocks in this file sorted.

Cargo.toml
```diff
@@ -124,7 +130,12 @@ sha3 = "0.10.6"
...
-tracing-subscriber = "0.3.16"
+tracing = "0.1.26"
+tracing-subscriber = { version = "0.3.17" }
+tracing-appender = "0.2.2"
 url = { version = "2.3.1", features = ["serde"] }
 urlencoding = "2.1.2"
+fern = "0.6.2"
+log = "0.4.17"
+chrono = "0.4.24"
+flate2 = "1.0.26"
```
Indy2222, line 141: Please make sure that all text based files end with a new line.

src/main.rs
```diff
@@ -16,56 +17,99 @@ use de_gui::GuiPluginGroup;
...
+                .add_directive("async_io=info".parse().unwrap())
+                .add_directive("polling=debug".parse().unwrap())
+                .add_directive("wgpu_hal=info".parse().unwrap())
+                .add_directive("naga=info".parse().unwrap()),
+        )
+        .with(fmt::layer().with_writer(std::io::stdout))
+        .with(fmt::layer().with_writer(non_blocking_log_writer));
+    tracing::subscriber::set_global_default(collector).expect("Unable to set a global collector");
+
+    info!("Logger initialized");
```
Indy2222, line 59: I think that this log has little added value so I would either remove it (preferably) or reduce the severity. Just one line below that is another log about the game starting.

src/main.rs
```diff
@@ -16,56 +17,99 @@ use de_gui::GuiPluginGroup;
...
-        .add_plugins(SignsPluginGroup)
-        .add_plugins(SpawnerPluginGroup)
-        .add_plugins(MovementPluginGroup)
-        .add_plugins(ControllerPluginGroup)
-        .add_plugins(CameraPluginGroup)
-        .add_plugins(BehaviourPluginGroup)
-        .add_plugins(CombatPluginGroup)
-        .add_plugins(ConstructionPluginGroup);
+    // for file name
+    let dt = chrono::Local::now();
```
Indy2222, line 37: I would move all this logic to `de_log`.

Note that the log file is passed to the logging plugin in an unnecessary complex manner (public property of the plugin group, then the plugin then a resource). None of that would be needed if the setup is done in a startup plugin.

crates/log/src/setup.rs
```diff
@@ -0,0 +1,215 @@
...
+    fn build(&self, app: &mut App) {
+        app.insert_resource(CurrentLogFile {
+            path: self.current_log_file.clone(),
+        });
+        app.add_startup_system(compress_old_logs);
+    }
+}
+
+// get all log files in the /logs directory that are not the current log file then compress them and delete the originals
+fn compress_old_logs(current_log_file: Res<CurrentLogFile>) {
```
Indy2222, line 26: How quickly to logs grow (e.g. KiB per minute of play time)? I would not implement the compression unless the logs get very large soon. And even in such a case, reducing some log levels might be a better solution.

Anyways, since this change is controversial, I would remove it from this PR. If we agree to do it in any form, you can open a separate follow up PR. That way, the rest will get merged sooner.

Indy2222 (changes requested):
I am happy to see this effort. Please let me know if you need any help or want to discuss anything on Discord.

Indy2222:
> I think some more discussion may be needed around log filtering and weather we want a really verbose log for crashes. I have tried to make it easy to just subscribe new writers onto the tracing log. It is a little bit against the overall layout of the code-base to put so much code into the main.rs, do we need to move it?

Yes, I would move most of the logging setup code to de_log crate.

crates/log/src/setup.rs
JackCrumpLeys: Game startup consumes about 1MB and after playing for a while it reaches about 2MB. The compression is able to reduce file size from 1MB to around 50kb in only 2ms. 1 MB is not much but after playing hour long games with a lot going on in them I'd imagine it would become quite big. Gzip is a very low-cost compression algorithm.

I will remove this from the PR but I do think that compression is a good thing to have further discussion around in the future.

JackCrumpLeys:
I think I have resolved all of those issues, my fork will run the actions I'll wait on those but I think its all good. 
de_log probably does not need to be a proper plugin just be passed app so it can insert the guard. For the future I do think that it would be good to have a very verbose log just for crash stuff and then a normal verbosity log for general things. That is why I have used the layer pattern to build the logger because using layers you can add writers with different directives for more complicated logging in the future.

any further comments?

crates/log/src/setup.rs
```diff
@@ -0,0 +1,44 @@
+use bevy::prelude::*;
+use tracing::Level;
+use tracing_appender::non_blocking::WorkerGuard;
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::{fmt, EnvFilter};
+
+pub(crate) struct LogPlugin {}
```
Indy2222, line 7: I would remove the `{}` and make this unit-like struct.

crates/log/src/setup.rs
```diff
@@ -0,0 +1,44 @@
...
+    #[allow(dead_code)]
+    guard: WorkerGuard,
+}
+
+impl Plugin for LogPlugin {
+    fn build(&self, app: &mut App) {
+        // for file name
+        let dt = chrono::Local::now();
+
+        let file_name = dt.format("%Y-%m-%d %H:%M:%S.log").to_string();
```
Indy2222, line 21: * I would change the file name so it does not contain a space (to make it easier to work with them in the command line) or a colon (this causes issues on MacOS as I was told).

* Opinionated nitpick: I would make `file_name` a `PathBuf` so that the trick `&*file_name` is not needed: `let path: PathBuf = dt.format("%Y-%m-%dT%H%M%S.log").to_string().into();`

Indy2222 (changes requested):

crates/log/src/setup.rs
JackCrumpLeys: Would this format be good? %Y-%m-%d_%H-%M-%S.log (2023-05-09_10-13-17.log)

Indy2222:
Superseded by #447 (I could not push to this branch).

PR #453, opened 2023-05-25
Add conveinience macros to make adding and reading config easier

JackCrumpLeys:
The means of creating and editing configuration is easier to do. The simplest config can be created by deriving default:
```rs
#[derive(Config, Default)] // deriving config adds the checks and the partial struct
pub struct TestConfig {
    #[is_finite] // automatic implementation of finite check for all fields

    #[ensure(foo <= bar, "foo")] // checks have access to all named fields in the struct
    #[ensure(foo <= baz, "bar")] // multiple checks
    pub foo: u32,
    #[check(|v: u32| Ok(ensure!(*v > 0, "foo")))] // we can use a closure for complicated code (for better type checking you can write a function outside of the closure and call it within)
    pub bar: u32,
    pub baz: u32,
} 

// ...


bundle_config!(
    // final_name:from:to
    test:TestConfig:TestConfig, // from and to are the same because it is basic data and the deserilized formats are good
)
``` 
`bundle_config!` bundles configuration neatly into a single struct.

the most important part is the generated `load` method which loads the
configuration from a file.

It also manages converting into final desired data structure.

- `name` is the name of the field in the Configuration struct
- `type_from` is the implementer of Config
- `type_into` is the final desired data structure must be able to fulfill the TryInto trait.

I am looking forward to feedback as this is highly opinionated. Also note this is my first experience with rust derive macro so I may not be doing this the best way. I had a lot of fun making the macro and I would not be too sad if you decide that this is not ergonomic enough.


crates/conf/src/macros.rs
```diff
@@ -0,0 +1,142 @@
...
+                            let check = value.check();
+                            if let Err(err) = check {
+                                for err in err {
+                                    error!("Failed check: {:?}", err);
+                                    errors.push(err);
+                                }
+                            }
+                        )*
+                        if !errors.is_empty() {
+                            error!("Failed checks:");
```
Indy2222, line 101: Ditto, I would change this to trace or remove it whatsoever. `ConfigLoadError::CheckErrors` should be comprehensively handled in `plugin.rs` after the loading task returns.

crates/conf/src/macros.rs
```diff
@@ -0,0 +1,142 @@
...
+        #[derive(Deserialize, Debug, Clone, Default)]
+        struct PartialConfiguration {
+            $(
+                $name: Option<paste! {[<Partial $type_from>]}>,
+            )*
+        }
+
+        impl Configuration {
+            $(
+                pub fn $name(&self) -> &$type_into {
```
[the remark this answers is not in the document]
JackCrumpLeys: this method is only for returning Config structs from the Configuration object I don't foresee those being copy.

crates/conf/src/plugin.rs
```diff
@@ -25,16 +28,23 @@ impl Plugin for ConfPlugin {
...
 
 fn cleanup(mut commands: Commands) {
     commands.remove_resource::<LoadingTask>();
 }
 
 fn start_loading(mut commands: Commands) {
     let task = IoTaskPool::get().spawn(async {
-        let path = conf_dir()?.join("conf.yaml");
-        load_conf(path.as_path()).await
+        let Some(base_conf_dir) = dirs::config_dir() else {
```
Indy2222, line 39: What is the reason for this change? 

Note that `thiserror` Crate has `#[from]` attribute which we can utilize to convert errors from `conf_dir()`. I think it is desirable to utilize functions from `de_core::fs` to stay DRY and keep it organized. A variant of the following might suffice:

```rust
let path = conf_dir().map_err(ConfigLoadError::from)?.join("conf.yaml");
Configuration::load(path.as_path()).await
```

crates/conf/src/plugin.rs
```diff
@@ -60,8 +70,18 @@ fn poll_conf(
                     toasts.send(ToastEvent::new(format!(
                         "Configuration loading failed: {err}"
                     )));
-                    let error: &dyn Error = err.as_ref();
-                    log_full_error!(error);
+                    let error: ConfigLoadError = err;
```
Indy2222, line 73: Why not just `match err` without this line?

crates/conf/src/plugin.rs
```diff
@@ -60,8 +70,18 @@ fn poll_conf(
...
+
+                    match error {
+                        ConfigLoadError::Error(e) => {
+                            error!("Config loading failed: {:?}", e)
+                        }
+                        ConfigLoadError::CheckErrors(e) => {
+                            for err in e {
```
Indy2222, line 80: I would improve the log by giving it a header + unifying multiple logs into a single one.

```rust
let mut error_log = String::from("Configuration validation errors:");
for err in errors {
    error_log.push('\n');
    error_log.push_str(&format!("{err}"));
}
```

crates/conf/src/plugin.rs
```diff
@@ -60,8 +70,18 @@ fn poll_conf(
...
                     )));
-                    let error: &dyn Error = err.as_ref();
-                    log_full_error!(error);
+                    let error: ConfigLoadError = err;
+
+                    match error {
+                        ConfigLoadError::Error(e) => {
+                            error!("Config loading failed: {:?}", e)
+                        }
+                        ConfigLoadError::CheckErrors(e) => {
```
Indy2222, line 79: `e` -> `errors` would make the code easier to read.

crates/conf/src/macros.rs
```diff
@@ -0,0 +1,146 @@
+use anyhow::Error;
+use thiserror::Error as ErrorDerive;
+
+#[derive(Debug, ErrorDerive)]
+pub enum ConfigLoadError {
+    #[error("Failed to load configuration: {0}")]
+    Error(Error),
+    #[error("Configuration check failed: {0:?}")]
+    CheckErrors(Vec<(String, Error)>),
```
Indy2222, line 9: I would enforce a specific type for each error. It would make it simpler. Ideally the type should implement Display (so we do not have to use debug printing). Maybe a String?

Indy2222 (changes requested):
Another batch of review comments. I still haven't fully reviewed the procedural macro but I have quickly read it.

crates/conf/macros/src/lib.rs
```diff
@@ -0,0 +1,217 @@
...
+                        check_fn = Some(quote! {
+                            (|#field_name: #arg_type| {
+                                ensure!(#condition, #string);
+                                Ok(())
+                            })
+                        });
+
+                        inner_check_fns.push(check_fn)
+                    } else if attr.path.is_ident("is_finite") {
+                        is_finite_check = !is_finite_check;
```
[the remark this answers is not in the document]
JackCrumpLeys: So that it can be toggled.

crates/conf/src/macros.rs
JackCrumpLeys: What do you mean by this? types for `CheckErrors` or just more `Error` types in the enum?

Indy2222: I mean that as of now, there is no convincing reason (AFAIK) to support generic field errors. So I would change this to `CheckErrors(Vec<(String, String)>)` and document that this is a tuples with field and error description respectively.

crates/conf/macros/src/lib.rs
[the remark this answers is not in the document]
Indy2222: I don't understand the toggling. Why isn't handle this attribute independently for each field?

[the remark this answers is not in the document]
JackCrumpLeys: In a lot of configs I foresee that there are many number values to be validated. So that it is not tedious to type out validation for each field. It's toggle so if you want to turn it on for a group of numbers then turn it off.  Have a look at the tests I wrote to see this behaviour.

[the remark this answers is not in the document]
JackCrumpLeys: In a lot of configs I foresee that there are many number values to be validated. So that it is not tedious to type out validation for each field. It's toggle so if you want to turn it on for a group of numbers then turn it off.  Have a look at the tests I wrote to see this behaviour.

crates/conf/macros/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+
+use itertools::Itertools;
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use quote::{quote, ToTokens};
+use syn::parse::{Parse, ParseStream};
+use syn::{parse_macro_input, Data, DeriveInput, Fields, Pat, Token};
+
+struct EnsureArgs(syn::Expr, syn::LitStr);
+impl Parse for EnsureArgs {
```
Indy2222, line 12: Please add an empty line above this one.

crates/conf/macros/tests/fails/bad_check_closure_parse.rs
```diff
@@ -0,0 +1,11 @@
...
+
+#[derive(Config)]
+pub struct TestConfig {
+    pub field1: u32,
+    #[check(foobar)]
+    pub field2: u32,
+    pub field3: u32,
+}
+
+fn main() {}
```
Indy2222, line 11: Please add a new line here and to every text-base file which does not already end with a new line character.

crates/conf/src/conf.rs
```diff
@@ -1,78 +1,88 @@
...
 
-use crate::persisted::{self, *};
+use crate::bundle_config;
 
-/// Top-level game configuration object.
-#[derive(Resource)]
-pub struct Configuration {
-    multiplayer: MultiplayerConf,
-    camera: CameraConf,
+// ---- Config structs hold deserialized and validated data before further processing or packaging into Configuration ----
```
Indy2222, line 14: This line seems too long.

crates/conf/macros/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+                            return syn::Error::new_spanned(
+                                closure.inputs,
+                                "expected a closure with one argument",
+                            )
+                            .to_compile_error()
+                            .into();
+                        }
+
+                        let arg = &closure.inputs[0];
+                        let arg = match arg {
```
Indy2222, line 82: Optional nitpick: `let Pat::Type(arg) = arg else { return error here };`

crates/conf/macros/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+                                #body
+                            })
+                        });
+
+                        inner_check_fns.push(check_fn)
+                    } else if attr.path.is_ident("ensure") {
+                        // the ensure attribute has 2 inputs like this: #[ensure(move_margin > 0., "`move_margin` must be positive.")]
+                        // it uses the `ensure!()` macro from the anyhow crate
+
+                        // we need to get the two inputs separated by a comma separately. the first one is a boolean expression and the second one is a string literal
```
Indy2222, line 111: Ditto, this line (and a few other lines) seems a bit two long. Note that rustfmt set comment width to 80 by default (https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#comment_width).

crates/conf/macros/src/lib.rs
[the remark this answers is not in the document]
Indy2222: Now I understand, but I would go the path of least surprise. Id est I would handle this independently for each field. Or is this pattern widely used / known?

crates/conf/macros/src/lib.rs
```diff
@@ -0,0 +1,261 @@
...
+                let field_name = &field_info.field_name;
+                let field_type = &field_info.field_type;
+
+                let inner_check_fns = &mut field_info.check_fns;
+
+                let mut check_fn;
+                for attr in &field.attrs {
+                    // get the check attribute with closures like this: #[check(|v| v > 0)]
+                    if attr.path.is_ident("check") {
+                        let closure = match attr.parse_args::<syn::ExprClosure>() {
```
Indy2222, line 59: I would separate each branch (`check`, `ensure` and `is_finite`) to a function so that the high level structure of this macro is easier to read.

Indy2222 (changes requested):

JackCrumpLeys:
Sorry new to rebasing

JackCrumpLeys:
OK I did it

Indy2222 (approved):

PR #477, opened 2023-06-07
Show updated health bars for a short time

Zakru:
Implements #388

The bar is set visible using an event, and adds a timer component or updates an existing one. Once the timer expires, the visibility flag is unset and the timer removed.

Feel free to suggest changes, I wrote this very spontaneously.

### DIFF
--- a/crates/signs/src/bars.rs
+++ b/crates/signs/src/bars.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use bevy::{
     pbr::{MaterialPipeline, MaterialPipelineKey, NotShadowCaster, NotShadowReceiver},
     prelude::*,
@@ -19,11 +21,14 @@ use de_core::{
 };
 use de_objects::SolidObjects;
 
-use crate::{DISTANCE_FLAG_BIT, MAX_VISIBILITY_DISTANCE};
+use crate::{DISTANCE_FLAG_BIT, MAX_VISIBILITY_DISTANCE, UPDATE_TIMER_FLAG_BIT};
 
 /// Vertical distance in meters between the bar center and the top of the
 /// parent entity collider.
 const BAR_HEIGHT: f32 = 2.;
+
+/// Duration that a bar is visible when its value is updated.
+const UPDATE_VISIBILITY_DURATION: Duration = Duration::from_secs(3);
 
 const ATTRIBUTE_POSITION: MeshVertexAttribute =
     MeshVertexAttribute::new("Position", 732918835, VertexFormat::Float32x2);
@@ -59,6 +64,12 @@ impl Plugin for BarsPlugin {
                     .run_if(in_state(AppState::InGame))
                     .before(VisibilitySet::Update)
                     .after(DistanceSet::Update),
+            )
+            .add_system(
+                update_visibility_timer
+                    .in_base_set(GameSet::PostUpdate)
+                    .run_if(in_state(AppState::InGame))
+                    .before(VisibilitySet::Update),
             );
     }
 }
@@ -182,6 +193,18 @@ impl Material for BarMaterial {
 #[derive(Component)]
 struct BarChild(Entity);
 
+#[derive(Component)]
+struct BarUpdateTimer(Timer);
+
+impl Default for BarUpdateTimer {
+    fn default() -> Self {
+        let mut timer = Timer::new(UPDATE_VISIBILITY_DURATION, TimerMode::Once);
+        // Avoid triggering visibility after spawning.
+        timer.tick(UPDATE_VISIBILITY_DURATION);
+        Self(timer)
+    }
+}
+
 fn setup(mut commans: Commands, mut meshes: ResMut<Assets<Mesh>>) {
     commans.insert_resource(BarMesh(meshes.add(bar_mesh(1.5, 0.3))));
 }
@@ -215,6 +238,7 @@ fn spawn(
                 NotShadowCaster,
                 NotShadowReceiver,
                 VisibilityFlags::default(),
+                BarUpdateTimer::default(),
             ))
             .id();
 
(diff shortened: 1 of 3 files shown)


crates/signs/src/bars.rs
```diff
@@ -269,6 +293,23 @@ fn update_visibility_distance(
     }
 }
 
+fn update_visibility_timer(
```
Indy2222, line 296: I think that the implementation might be a bit simplified and made more efficient with the following changes to this commit:

- Change `spawn()` system so that each bar has `BarUpdateTimer` component.
- Change `update_visibility_timer` so that 
  - it iterates over all `BarUpdateTimer` components and changes visibility flag `TIMER_BAR_ID` to `false` if the timer [just_finished](https://docs.rs/bevy/latest/bevy/prelude/struct.Timer.html#method.just_finished) (among other things to avoid unnecessary change detection).
  - before calling `tick()` it (conditionally to avoid change detection) changes `TIMER_BAR_ID` to `true` if `elapsed()` time of the timer is 0
- Change `update_value()` system so that it does not insert `BarUpdateTimer` component (see first point, we can avoid frequent re-adding by keeping the component forever). It should only reset the timer.

crates/signs/src/lib.rs
```diff
@@ -13,6 +13,8 @@ mod pole;
 const MAX_VISIBILITY_DISTANCE: f32 = 140.;
 const DISTANCE_FLAG_BIT: u32 = 0;
 
+const TIMER_BAR_ID: u32 = 2;
```
Indy2222, line 16: I would rename `TIMER_BAR_ID` to `UPDATE_TIMER_FLAG_BIT` to so that the naming is consistent and more suggestive.

crates/signs/src/lib.rs
```diff
@@ -13,6 +13,8 @@ mod pole;
 const MAX_VISIBILITY_DISTANCE: f32 = 140.;
 const DISTANCE_FLAG_BIT: u32 = 0;
```
Indy2222, line 15: Opinionated: I would remove this empty line here.

crates/signs/src/bars.rs
```diff
@@ -182,6 +188,15 @@ impl Material for BarMaterial {
 #[derive(Component)]
 struct BarChild(Entity);
 
+#[derive(Component)]
+struct BarVisibilityTimer(Timer);
+
+impl Default for BarVisibilityTimer {
+    fn default() -> Self {
+        Self(Timer::from_seconds(3., TimerMode::Once))
```
Indy2222, line 196: I would use a constant `UPDATE_VISIBILITY_DURATION` (defined near the top of the file) and change this to `Timer::new(UPDATE_VISIBILITY_DURATION, TimerMode::Once)`.

Indy2222 (changes requested):
It is cool to see this implemented! ❤️

It is good from high level perspective. I made some lower level improvement suggestions.

crates/signs/src/bars.rs
Zakru: I figured that might be the case, just decided to do the first approach that came to mind :)

crates/signs/src/lib.rs
Zakru: Ok. Fwiw I took that scheme from `controller/lib.rs`, just didn't notice we already had related constants here.

crates/signs/src/bars.rs
```diff
@@ -19,12 +19,15 @@ use de_core::{
...
 
-use crate::{DISTANCE_FLAG_BIT, MAX_VISIBILITY_DISTANCE};
+use crate::{DISTANCE_FLAG_BIT, MAX_VISIBILITY_DISTANCE, UPDATE_TIMER_FLAG_BIT};
 
 /// Vertical distance in meters between the bar center and the top of the
 /// parent entity collider.
 const BAR_HEIGHT: f32 = 2.;
 
+/// Duration in seconds that a bar is visible when its value is updated.
+const UPDATE_VISIBILITY_DURATION: f32 = 3.;
```
Indy2222, line 29: Nitpick: We can use `Duration` like this `Duration::from_secs(3)`. It is a const function and this way, it would be harder to misinterpret the units.

crates/signs/src/bars.rs
```diff
@@ -269,6 +290,26 @@ fn update_visibility_distance(
     }
 }
 
+fn update_visibility_timer(
+    parents: Query<&BarChild, With<Active>>,
```
Indy2222, line 294: I think these are not needed, we can iterate directly over all `bars`. Or is there a reason for the indirection?

Indy2222 (changes requested):
Two minor comments and I think we can merge it.

crates/signs/src/bars.rs
Zakru: Ah, that would be me editing code too quickly at 1 AM.

crates/signs/src/bars.rs
```diff
@@ -182,6 +193,15 @@ impl Material for BarMaterial {
 #[derive(Component)]
 struct BarChild(Entity);
 
+#[derive(Component)]
+struct BarUpdateTimer(Timer);
+
+impl Default for BarUpdateTimer {
+    fn default() -> Self {
+        Self(Timer::new(UPDATE_VISIBILITY_DURATION, TimerMode::Once))
```
Indy2222, line 201: I've realized during the testing that this implementation leads to all bars being briefly displayed after entity spawning. One hacky way to solve this is to do a variation of the following:

```rust
let mut timer = Timer::new(UPDATE_VISIBILITY_DURATION, TimerMode::Once);
// Avoid triggering visibility after spawning.
timer.tick(UPDATE_VISIBILITY_DURATION);
Self(timer)
```

Indy2222 (changes requested):

Indy2222 (approved):

Issue #36, opened 2022-04-06
Highlight Selected Entities


Indy2222:
- [x] Draw circles on terrain below selected units
  - Solved in #209
- [ ] Draw squares on terrain below selected buildings

Issue #145, opened 2022-06-27
Make it possible to rotate to be constructed buildings


Indy2222:
This is follow up to #23.

Players should have to possibility to choose rotation (along Y axis) of to be constructed buildings.

PR #482, opened 2023-06-08
Square markers for buildings

Zakru:
Completes the second subtask of #36. Prior discussion can be found in #480 

Adds square markers to highlight selected buildings.

Draft until the following is done:

- [x] Test performance
- [x] Improve marker size calculation
- [x] Squash?

### DIFF
--- a/assets/shaders/terrain.wgsl
+++ b/assets/shaders/terrain.wgsl
@@ -14,8 +14,9 @@
 const TEXTURE_SIZE = 16.;
 const SHAPE_COLOR = vec4<f32>(1., 1., 1., 0.75);
 const SHAPE_THICKNESS = 0.15;
-// Keep thie array lenght in sync with /crates/terrain/src/shader.rs.
+// Keep these array lengths in sync with /crates/terrain/src/shader.rs.
 const MAX_KD_TREE_SIZE = 127u;
+const MAX_RECTANGLE_ARRAY_SIZE = 31u;
 
 struct KdTreeNode {
     @align(16) location: vec2<f32>,
@@ -27,11 +28,23 @@ struct KdTree {
     count: u32,
 };
 
+struct Rectangle {
+    inverse_transform: mat3x3<f32>,
+    half_size: vec2<f32>,
+};
+
+struct Rectangles {
+    items: array<Rectangle, MAX_RECTANGLE_ARRAY_SIZE>,
+    count: u32,
+};
+
 @group(1) @binding(0)
 var<uniform> circles: KdTree;
 @group(1) @binding(1)
-var terrain_texture: texture_2d<f32>;
+var<uniform> rectangles: Rectangles;
 @group(1) @binding(2)
+var terrain_texture: texture_2d<f32>;
+@group(1) @binding(3)
 var terrain_sampler: sampler;
 
 struct FragmentInput {
@@ -146,6 +159,18 @@ fn draw_circles(base: vec4<f32>, uv: vec2<f32>) -> vec4<f32> {
     return output_color;
 }
 
+fn draw_rectangles(base: vec4<f32>, uv: vec2<f32>) -> vec4<f32> {
+    for (var i = 0u; i < rectangles.count; i++) {
+        let rectangle = rectangles.items[i];
+        let local_uv = (rectangle.inverse_transform * vec3(uv, 1.0)).xy;
+        if all(abs(local_uv) <= rectangle.half_size + SHAPE_THICKNESS) && any(rectangle.half_size <= abs(local_uv)) {
+            return mix_colors(base, SHAPE_COLOR);
+        }
+    }
+
+    return base;
+}
+
 @fragment
 fn fragment(in: FragmentInput) -> @location(0) vec4<f32> {
     var pbr_input: PbrInput = pbr_input_new();
@@ -195,5 +220,6 @@ fn fragment(in: FragmentInput) -> @location(0) vec4<f32> {
 #endif
 
     output_color = draw_circles(output_color, in.uv);
+    output_color = draw_rectangles(output_color, in.uv);
     return output_color;
 }
--- a/crates/controller/src/selection/bookkeeping.rs
+++ b/crates/controller/src/selection/bookkeeping.rs
@@ -2,7 +2,7 @@ use ahash::AHashSet;
 use bevy::{ecs::system::SystemParam, prelude::*};
 use de_core::{baseset::GameSet, gamestate::GameState};
 use de_signs::{UpdateBarVisibilityEvent, UpdatePoleVisibilityEvent};
-use de_terrain::CircleMarker;
+use de_terrain::MarkerVisibility;
 
 use crate::SELECTION_BAR_ID;
 
@@ -164,15 +164,13 @@ fn update_selection(mut events: EventReader<SelectEvent>, selector_builder: Sele
 
 fn selected_system(
     mut events: EventReader<SelectedEvent>,
-    mut markers: Query<&mut CircleMarker>,
+    mut markers: Query<&mut MarkerVisibility>,
     mut bars: EventWriter<UpdateBarVisibilityEvent>,
     mut poles: EventWriter<UpdatePoleVisibilityEvent>,
 ) {
     for event in events.iter() {
-        if let Ok(mut marker) = markers.get_mut(event.0) {
-            marker
-                .visibility_mut()
-                .update_visible(SELECTION_BAR_ID, true);
+        if let Ok(mut visibility) = markers.get_mut(event.0) {
+            visibility.0.update_visible(SELECTION_BAR_ID, true);
         }
 
         bars.send(UpdateBarVisibilityEvent::new(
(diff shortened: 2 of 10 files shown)


assets/shaders/terrain.wgsl
```diff
@@ -27,11 +28,23 @@ struct KdTree {
     count: u32,
 };
 
+struct Square {
```
Indy2222, line 31: This should be a rectangle which may be rotated (not axis-aligned). This is because the game already support non-rectangular buildings and rotated (around vertical axis) buildings.

You can test the rotation by modifying a map, it stores angle for each building. New building construction will also support rotation in the future (#145).

We might want to represent it as a 3x3 matrix (with rotation + translation) and half_size 2D vector so we can quickly project `uv` point to the local coordinates of the rectangle (`projected = proj * vec3(uv, 1)`) and just test that `abs(projected) < half_size`. This way, we would need to pre-compute this for every building only when spawned (or more robustly, its `Transform` changed) which is way better than doing more calulations for every rendered pixel during every rendered frame.

assets/shaders/terrain.wgsl
```diff
@@ -14,8 +14,9 @@
 const TEXTURE_SIZE = 16.;
 const SHAPE_COLOR = vec4<f32>(1., 1., 1., 0.75);
 const SHAPE_THICKNESS = 0.15;
-// Keep thie array lenght in sync with /crates/terrain/src/shader.rs.
+// Keep these array lengths in sync with /crates/terrain/src/shader.rs.
 const MAX_KD_TREE_SIZE = 127u;
+const MAX_SQUARE_ARRAY_SIZE = 31u;
```
Indy2222, line 19: Make sure this is small enough until we fix #216. Hopefully, once #216 is resolved, we will be able to remove this limitation and draw the selection boxes / circles around every single selected entity.

crates/controller/src/selection/bookkeeping.rs
```diff
@@ -164,15 +164,13 @@ fn update_selection(mut events: EventReader<SelectEvent>, selector_builder: Sele
...
+    mut markers: Query<&mut MarkerVisibility>,
     mut bars: EventWriter<UpdateBarVisibilityEvent>,
     mut poles: EventWriter<UpdatePoleVisibilityEvent>,
 ) {
     for event in events.iter() {
-        if let Ok(mut marker) = markers.get_mut(event.0) {
-            marker
-                .visibility_mut()
-                .update_visible(SELECTION_BAR_ID, true);
+        if let Ok(mut visibility) = markers.get_mut(event.0) {
```
JackCrumpLeys, line 172: What do we do if not Ok? Or is this something that should never happen?

Indy2222 (changes requested):
Cool! This was a major missing piece in the UI.

We need to generalize the marker a bit and make sure the performance is sufficient. Otherwise it looks very good!

crates/controller/src/selection/bookkeeping.rs
```diff
@@ -187,15 +185,13 @@ fn selected_system(
...
+    mut markers: Query<&mut MarkerVisibility>,
     mut bars: EventWriter<UpdateBarVisibilityEvent>,
     mut poles: EventWriter<UpdatePoleVisibilityEvent>,
 ) {
     for event in events.iter() {
-        if let Ok(mut marker) = markers.get_mut(event.0) {
-            marker
-                .visibility_mut()
-                .update_visible(SELECTION_BAR_ID, false);
+        if let Ok(mut visibility) = markers.get_mut(event.0) {
```
JackCrumpLeys, line 193: Ditto what do we do on error?
Could do something like log it using `warn!()` or `error!()` at least.

JackCrumpLeys (commented):
Had a skim, some minor changes to error handling. I would like there to be some tests before merge but no pressure at this point.

JackCrumpLeys:
this is working towards #36 (just wanted to mention it)

Zakru:
Tested performance, can't observe a single FPS's difference on my machine before and after, with more than the max buildings selected.

Both tests running with `testing` profile, mainly hovering in the 142.x FPS range, running on RTX 2060 + i5 4690K so a relatively outdated CPU

crates/controller/src/selection/bookkeeping.rs
Indy2222: This is because `SelectedEvent` for a particular entity might have been sent during an earlier scheduling base set. Therefore, other systems might get a chance to despawn the entity.

JackCrumpLeys:
> Tested performance, can't observe a single FPS's difference on my machine before and after, with more than the max buildings selected.
> 
> Both tests running with `testing` profile, mainly hovering in the 142.x FPS range, running on RTX 2060 + i5 4690K

What platform are you running on?

crates/controller/src/selection/bookkeeping.rs
Indy2222: Please see my other comment. We don't want to make strong assumptions about timing of when the event was sent (or the event leading to this event) and about other systems not despawning the entity in the meantime. I would not even log a warning. Maybe a trace!?

JackCrumpLeys: Ok, I was just wondering.

Zakru:
Running on Windows 10

Zakru:
Performance test wasn't good actually, I can see much higher FPS when the window is unfocused, so there must be some magic cap by NVIDIA that gets activated for some reason. Need to investigate and re-test.

Zakru:
New results:

| FPS when | Nothing selected | Max+ buildings selected |
|-|-|-|
| Plain square impl | ~450 | ~410 |
| Matrix impl | ~450 | ~400 |

assets/shaders/terrain.wgsl
Zakru: I made a commit of what I think that would look like, although it still ends up calculating a bunch of things per-pixel so I feel I might have missed a point somewhere

crates/core/src/projection.rs
```diff
@@ -68,6 +68,16 @@ impl ToFlat<Aabb2D> for Aabb3D {
     }
 }
 
+impl ToFlat<Mat3> for Mat4 {
+    fn to_flat(self) -> Mat3 {
+        Mat3::from_cols_array_2d(&[
+            [self.x_axis.x, self.x_axis.z, 0.],
+            [-self.z_axis.x, -self.z_axis.z, 0.],
+            [self.w_axis.x, self.w_axis.z, 1.],
```
Indy2222, line 76: Isn't `z` inverted here?

crates/spawner/src/spawner.rs
```diff
@@ -83,6 +85,13 @@ fn spawn(
                 match active_type {
                     ActiveObjectType::Building(_) => {
                         entity_commands.insert(StaticSolid);
+
+                        let half_size: Vec2 =
+                            solid.ichnography().local_aabb().half_extents().into();
+                        entity_commands.insert(RectangleMarker::new(
```
Indy2222, line 91: Wouldn't the following be a bit nicer API?

```rust
let local_aabb = solid.ichnography().local_aabb();
RectangleMarker::from_aabb_transform(local_aabb, transform)
```

Among other things, it would remove the necessity to define/import `RECTANGLE_MARKER_MARGIN` here.

Indy2222 (changes requested):
I have only a few nitpicks. 

Could you squash the commits and add `Fixes #36` to the commit description?

crates/core/src/projection.rs
Indy2222: It is not, I was just confused.

assets/shaders/terrain.wgsl
```diff
@@ -146,6 +159,22 @@ fn draw_circles(base: vec4<f32>, uv: vec2<f32>) -> vec4<f32> {
     return output_color;
 }
 
+fn draw_rectangles(base: vec4<f32>, uv: vec2<f32>) -> vec4<f32> {
+    var paint = false;
+    for (var i = 0u; i < rectangles.count; i++) {
+        let rectangle = rectangles.items[i];
+        let local_uv = (rectangle.inverse_transform * vec3(uv, 1.0)).xy;
+        if all(abs(local_uv) <= rectangle.half_size + SHAPE_THICKNESS) && any(rectangle.half_size <= abs(local_uv)) {
```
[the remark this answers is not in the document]
Zakru: Actually no. The X and Y parts form *inverse* strips along the opposite axis, so they need to be ORd. See helpful infographic below :D
X and Y parts of the condition

assets/shaders/terrain.wgsl
```diff
@@ -146,6 +159,22 @@ fn draw_circles(base: vec4<f32>, uv: vec2<f32>) -> vec4<f32> {
...
+            paint = true;
```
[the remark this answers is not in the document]
Zakru: Hmm, I didn't like this logic too much either. Would just like to check, does that kind of conditional return affect the shader's performance?

assets/shaders/terrain.wgsl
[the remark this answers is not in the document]
Indy2222: Your are right, my bad!

assets/shaders/terrain.wgsl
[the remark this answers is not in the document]
Indy2222: I don't know what it compiles to, but I would hope that we can spare a few iterations inside the lines. But since the lines take only a small fraction of the screen space, it will have a negligible impact.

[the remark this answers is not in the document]
Zakru: Roger

Zakru:
Ninja `fmt`

Indy2222 (approved):

Issue #473, opened 2023-06-07
Simple Batteries


Indy2222:
As a step towards initial support for energy in the game, lets implement batteries:

- [x] Every unit & building has a battery. The battery has maximum capacity in Joules and current energy level in Joules.
- [x] The battery discharges at a constant rate (a constant in the code).
- [ ] Movement of unit is blocked once the energy goes to 0.
- [ ] Attacking is blocked once the energy goes below X (a constant in the code).
- [ ] Manufacturing is blocked once the energy goes below Y (a constant in the code).

Note that charging is not handled as part of this issue.

PR #483, opened 2023-06-09
Implement simple batteries

JackCrumpLeys:
This implements the first step towards energy (#473).


crates/energy/src/battery/systems.rs
```diff
@@ -0,0 +1,23 @@
+use bevy::prelude::*;
```
Indy2222, line 1: I would merge `systems.rs`, `component.rs` and `mod.rs` to a single `battery.rs`. We can split it once it grows larger.

There are two reasons for this:

* Currently, the combine length of the modules is below 200 lines and the modules are not super complex. We can always split it later once the code size / complexity grows.
* Thorough the codebase, plugins, their systems and related components are kept together. I prefer that we split the plugin to smaller ones to keep the modules manageable (once we reach that point).

crates/energy/src/battery/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+
+pub(crate) struct BatteryPlugin;
+
+impl Plugin for BatteryPlugin {
+    fn build(&self, app: &mut App) {
+        app.add_system(systems::discharge_battery);
+    }
+}
+
+/// The rate at which the battery discharges in kiloJoules per second.
```
Indy2222, line 14: I would use base units from SI (we can convert in UI) unless there is a technical reason not to do so (e.g. fixed point arthritics).

crates/energy/src/battery/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+const DISCHARGE_RATE: f32 = 30.;
+/// The minimum energy level in kiloJoules at which the unit can still move.
+const MIN_MOVE_ENERGY: f32 = 74.; // based on how many a car uses per second
+/// The minimum energy level in kiloJoules at which the unit can still attack.
+const MIN_ATTACK_ENERGY: f32 = 20_000.; // google said that a rail-gun takes 25 Mj
```
Indy2222, line 19: There are two different concepts which are being misleadingly merge to a single one here:

* Minimum required energy for some kind of activity to keep happening. There are several reasons for this mechanism, the main one is that the vehicle does not loose basic functionality due to a death battery. Once below below e.g. 20% or some Joules, it is time to use energy wisely (i.e. movement only).
* Energy physically required to perform an action, e.g. charge needed to fire a laser gun.

Also, drones are currently equipped with lasers not rail-guns so the comment is a bit misleading. Not that the actual value matter much at this point during the development.

crates/energy/src/battery/component.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub fn charge(&mut self, energy: f32) {
+        debug_assert!(energy.is_finite());
+        debug_assert!(energy >= 0.);
+        debug_assert!(energy <= self.capacity - self.energy);
+
+        self.energy += energy;
+    }
+
+    /// Get percentage of energy remaining in the battery.
+    pub fn energy_percentage(&self) -> f32 {
```
Indy2222, line 109: I would call it fraction not percentage since the value is between 0 and 1. But I actually like that it is a fraction and not percentage.

Is it even used? Lets not introduce dead code.

crates/energy/src/battery/component.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        self.energy -= energy;
+    }
+
+    /// Tries to charge the battery by the given amount of energy.
+    ///
+    /// # Returns
+    ///
+    /// `true` if the battery was charged, `false` otherwise.
+    pub fn try_charge(&mut self, energy: f32) -> bool {
```
Indy2222, line 87: Ditto, please remove all methods which are not yet used.

crates/energy/src/battery/component.rs
```diff
@@ -0,0 +1,112 @@
+use bevy::prelude::*;
+
+use crate::battery::{MIN_ATTACK_ENERGY, MIN_MOVE_ENERGY};
+
+/// The battery component is used to store the energy level of an entity.
+///
+/// # fields
```
Indy2222, line 7: Rust's standard way of documenting arguments used throughout the repository is this:

```rust
/// # Arguments
///
/// * `capacity` - ...
///
/// * `energy` - ...
```

I.e. different title, empty lines and stars for bullet points.

crates/energy/src/battery/component.rs
```diff
@@ -0,0 +1,112 @@
...
+/// - `capacity`: The maximum capacity of the battery in kilojoules.
+/// - `energy`: The current energy level of the battery in kilojoules.
+#[derive(Component, Debug, Clone, Copy, PartialEq, PartialOrd)]
+pub struct Battery {
+    capacity: f32, // in kilojoules
```
Indy2222, line 12: Maybe `capacity_joules` would be a less error prone name? What do you think?

crates/energy/src/battery/systems.rs
```diff
@@ -0,0 +1,23 @@
...
+/// Discharges the battery of a unit.
+///
+/// # Parameters
+///
+/// - `time`: The time since the last update in seconds.
+/// - `battery`: The battery.
+pub(crate) fn discharge_battery(time: Res<Time>, mut battery: Query<&mut Battery>) {
+    for mut battery in battery.iter_mut() {
+        let energy = battery.energy();
+        if energy == 0. {
```
Indy2222, line 15: `if battery.energy() == 0.` would be more concise.

crates/energy/src/battery/component.rs
```diff
@@ -0,0 +1,112 @@
...
+        if energy >= self.energy {
+            return false;
+        }
+
+        self.discharge(energy);
+        true
+    }
+
+    /// Directly discharges the battery by the given amount of energy.
+    pub fn discharge(&mut self, energy: f32) {
```
Indy2222, line 74: * What about unifying (dis)charging to a single method? E.g. `change(&mut self, delta: f32)`?
* I would change the implementation to `self.energy = (self.energy + delta).clamp(0., self.capacity);`. Currently, one cannot go to exact zero unless the update coincidentally hits exactly 0.

Indy2222 (changes requested):
Thank you for the contribution!

I had a bunch of comments but the overall structure is good.

crates/energy/src/battery/component.rs
JackCrumpLeys: I have plans to use these pieces of code (Keep in mind this is still a draft). I plan to add a basic UI to keep track of energy as part of this pr. At the most basic level just a key triggered debug screen.

crates/energy/src/battery/systems.rs
JackCrumpLeys: Can do, remnants of debugging prints lol.

crates/energy/src/battery/mod.rs
JackCrumpLeys: I think that we should just use the system where battery is depleted by preforming different actions. These can be loaded from object config. Do you think I should just go ahead and make that happen? the other thing seems like another feature/PR (sounds like something that should be configurable though the UI E.g at 20% stop firing at low threat targets + go to nearest safe power source)

crates/energy/src/battery/component.rs
JackCrumpLeys: How so?

Indy2222: Because it include units in the name. The downside is that it is longer. Also, we might start to enforce units at compile time once we get a decent UOM support. Therefore, I do not have a strong opinion on this.

JackCrumpLeys: Maybe to make it more readable I make an enum that takes different joule units as an input to the battery methods?

crates/energy/src/battery.rs
```diff
@@ -0,0 +1,203 @@
...
+/// The minimum energy level in kiloJoules at which the unit can still move.
+const MIN_MOVE_ENERGY: EnergyUnit = EnergyUnit::Kilojoules(74.); // based on how many a car uses per second
+/// The minimum energy level in kiloJoules at which the unit can still attack.
+const MIN_ATTACK_ENERGY: EnergyUnit = EnergyUnit::Megajoules(20.); // google said that a rail-gun takes 25 Mj
+/// The minimum energy level in kiloJoules at which factories can still produce units.
+const MIN_FACTORY_ENERGY: EnergyUnit = EnergyUnit::Megajoules(10.);
+/// The default capacity of the battery in kiloJoules.
+const DEFAULT_CAPACITY: EnergyUnit = EnergyUnit::Megajoules(100.); // 100 Mj
+
+pub enum EnergyUnit {
```
Indy2222, line 22: Please have a look at `de_uom`.

crates/energy/src/battery.rs
```diff
@@ -0,0 +1,203 @@
...
+impl Plugin for BatteryPlugin {
+    fn build(&self, app: &mut App) {
+        app.add_system(discharge_battery);
+    }
+}
+
+/// The rate at which the battery discharges in Joules per second.
+const DISCHARGE_RATE: f64 = 30_000.;
+/// The minimum energy level in kiloJoules at which the unit can still move.
+const MIN_MOVE_ENERGY: EnergyUnit = EnergyUnit::Kilojoules(74.); // based on how many a car uses per second
```
Indy2222, line 14: Why not 0? I think that it should be allowed to get to 0. for the most basic unit activity, i.e. movement.

crates/energy/src/battery.rs
```diff
@@ -0,0 +1,203 @@
...
+pub(crate) fn discharge_battery(time: Res<Time>, mut battery: Query<&mut Battery>) {
+    for mut battery in battery.iter_mut() {
+        let energy = battery.energy();
+        if energy == 0. {
+            continue;
+        }
+        let delta = time.delta_seconds();
+        let discharge = (DISCHARGE_RATE) * delta as f64;
+
+        battery.try_discharge(EnergyUnit::Joules(discharge));
```
Indy2222, line 164: I am not a fun of this back and forth movement between pure f32 and EnergyUnit. It looks overengineered. I tried something similar but more powerful in `de_uom` where i.e. multiplication works out of the box so you don't have to do this transitions. I am still undecided whether it brings more benefits than complexity.

Please have a look at `de_uom` and either use that or stick to f32.

crates/spawner/src/spawner.rs
```diff
@@ -72,6 +73,7 @@ fn spawn(
         match object_type {
             ObjectType::Active(active_type) => {
                 entity_commands.insert(Active);
+                entity_commands.insert(Battery::default()); // a bit of a placeholder as final solution is likely going to be in `SolidObject` configured though the object config
```
Indy2222, line 76: TL;DR: I would remove the comment.

Very likely the implementation will look like you suggest. However, we should keep our plans and ideas in GitHub issues & discussions. Plans tend to change over time and it is hard to keep all such comments up-to-date, well organized and searchable (e.g. via labels).

Indy2222 (changes requested):
Even though the PR is still a draft, I left a few comments. They mostly regard the style / taste in which things are implemented.

If you feel that such things are not well documented or communicated, please either let me know or open a PR with improvements to e.g. `CONTRIBUTING.md`. Thank you!

crates/energy/src/battery.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    /// Directly changes the energy level of the battery by the given amount of energy.
+    pub fn change(&mut self, delta: f64) {
+        debug_assert!(delta.is_finite());
+
+        self.energy = (self.energy + delta).clamp(0., self.capacity);
+    }
+
+    /// Get fraction of energy remaining in the battery. (a number between 0 and 1)
+    pub fn energy_percentage(&self) -> f64 {
```
Indy2222, line 111: I would rename this to `energy_fraction`

crates/energy/src/battery.rs
```diff
@@ -0,0 +1,170 @@
...
+/// - `time`: The time since the last update in seconds.
+/// - `battery`: The battery.
+pub(crate) fn discharge_battery(time: Res<Time>, mut battery: Query<&mut Battery>) {
+    for mut battery in battery.iter_mut() {
+        let energy = battery.energy();
+        if energy == 0. {
+            continue;
+        }
+        let delta = time.delta_seconds();
+        let discharge = (DISCHARGE_RATE) * delta as f64;
```
Indy2222, line 129: Why is there brackets around `DISCHARGE_RATE`?

crates/energy/src/battery.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    /// Get fraction of energy remaining in the battery. (a number between 0 and 1)
+    pub fn energy_percentage(&self) -> f64 {
+        self.energy / self.capacity
+    }
+}
+
+/// Discharges the battery of a unit.
+///
+/// # Parameters
```
Indy2222, line 118: Rename this to the standard `Arguments`

crates/energy/src/battery.rs
```diff
@@ -0,0 +1,170 @@
...
+///
+/// - `time`: The time since the last update in seconds.
```
Indy2222, line 120: * For consistency, please change this (and the item below) to ``- `time`: `` -> ``* `time` - ``
* Keep an empty line between individual bullet points.

crates/energy/src/battery.rs
```diff
@@ -0,0 +1,170 @@
...
+    fn default() -> Self {
+        Self {
+            capacity: DEFAULT_CAPACITY,
+            energy: DEFAULT_CAPACITY,
+        }
+    }
+}
+
+impl Battery {
+    pub fn new(capacity: f64, energy: f64) -> Self {
```
Indy2222, line 42: In this PR, I would remove `pub` from methods, structs, etc. where it is not necessary for compilation to succeed. Also, I would remove all unused methods and constants (to avoid merging dead code). We can implement action blocking (e.g. movement) in a follow up PR.

Indy2222 (changes requested):

crates/energy/src/battery.rs
JackCrumpLeys: If you want me to do it in another pr I could. Was gonna do this one. Would that be preferable?

JackCrumpLeys: Gonna do it in another one.

JackCrumpLeys:
In adding the UI I am not sure it I did it correctly / using best methods. I will move out of draft now. (I tried really hard to get a screenshot but on Ubuntu some issues with full screen disallow me to (I should open an issue for that))

crates/energy/src/battery.rs
```diff
@@ -0,0 +1,134 @@
...
+///
+/// * `battery` - The battery.
+pub(crate) fn discharge_battery(time: Res<Time>, mut battery: Query<&mut Battery>) {
+    for mut battery in battery.iter_mut() {
+        let energy = battery.energy();
+        if energy == 0. {
+            continue;
+        }
+
+        let delta = time.delta_seconds();
```
Indy2222, line 92: Nitpicks, we can move these two variable outside of the loop, no need to recompute it in every iteration.

crates/energy/src/battery.rs
```diff
@@ -0,0 +1,134 @@
...
+        let discharge = DISCHARGE_RATE * delta as f64;
+
+        battery.try_discharge(discharge);
```
Indy2222, line 95: I think that this system should discharge the battery all the way to 0. I.e. we can get rid of `try_discharge()` method and call `battery.change(energy_delta);` (note that we can rename `discharge` to `energy_delta` and make it negative).

Indy2222 (changes requested):
Could you separate the HUD modifications to another PR?

Otherwise, I had only two minor comments which we can quickly fix and merge it.

crates/controller/src/hud/details.rs
```diff
@@ -7,11 +7,11 @@ pub(crate) struct DetailsPlugin;
 
 impl Plugin for DetailsPlugin {
     fn build(&self, app: &mut App) {
-        app.add_system(spawn_details.in_schedule(OnEnter(GameState::Playing)));
+        app.add_system(setup.in_schedule(OnEnter(GameState::Playing)));
```
Indy2222, line 10: Can we remove these unrelated changes?

crates/energy/src/battery.rs
```diff
@@ -0,0 +1,117 @@
...
+/// * `battery` - The battery.
+pub(crate) fn discharge_battery(time: Res<Time>, mut battery: Query<&mut Battery>) {
+    let delta = time.delta_seconds();
+    for mut battery in battery.iter_mut() {
+        let energy = battery.energy();
+        if energy == 0. {
+            continue;
+        }
+
+        let discharge_delta = DISCHARGE_RATE * delta as f64;
```
Indy2222, line 76: Nitpick: We can move even this one out of the loop and make it negative: `let discharge_delta = -DISCHARGE_RATE * delta as f64;` so that we do not have to multiply by -1 in every iteration.

Indy2222 (changes requested):
Two nitpicks and we are good to go.

Indy2222 (approved):

Indy2222 (approved):

Issue #434, opened 2023-04-05
Display "arrows" from selected factories to delivery location


Indy2222:
This is a follow up of #428.

PR #484, opened 2023-06-09
Display line from factory to pole (part of #434)

0HyperCube:
Displays a line from the factory to the pole

TODO:
- [x] Texture line with some arrows
- [x] (maybe) animate the arrows
- [x] Cleanup transform handling


crates/signs/src/pole.rs
```diff
@@ -124,16 +141,23 @@ struct Pole {
...
-struct MovePoleEvent(Vec2, Vec2);
+struct MovePoleEvent {
+    old_location: Vec2,
+    pole_location: Vec2,
+    line_start: Vec2,
+}
 
 #[derive(Default, Resource)]
-struct Poles(AHashMap<Vec2Ord, Entity>);
+struct Poles(AHashMap<Vec2Ord, [Entity; 2]>);
```
Indy2222, line 160: This doesn't work correctly. You can reproduce the issue with:

1. open a game & build a second base
2. select the first & second base at once
3. right click on the map to place both poles to a single place

The reason why it doesn't work is pole-merging logic. I.e. when two pole owners place a pole to the same location, only a single pole is created.

IMO the best solution is the creation of a separate module `lines.rs` with the pole-independent line logic there. I would prefer such a solution because decoupling poles & lines would simplify the code, and it would give us more flexibility by allowing poles without lines and vice versa.

crates/signs/src/pole.rs
```diff
@@ -187,25 +211,38 @@ fn update_poles(
...
+        if !desired.contains_key(location) {
             to_despawn.push(location.0);
         }
     }
-    for location in &desired {
+    for (location, owner) in &desired {
+        let Ok(owner_transform) = transform_query.get(**owner) else {
+            continue;
+        };
+        let line_start = owner_transform.translation.to_flat();
```
Indy2222, line 233: The line should start at the spawn location of the owner factory.

assets/shaders/rally_point.wgsl
```diff
@@ -0,0 +1,27 @@
...
+fn fragment(
+    #import bevy_pbr::mesh_vertex_output
+) -> @location(0) vec4<f32> {
+    let scaled_x: f32 = uv.x * length(mesh.model[0].xyz);
+    let offset_y: f32 = abs(uv.y - 0.5) * material.pointiness;
+    let scaled_time: f32 = globals.time * material.speed;
+    let total_length = material.length + material.spacing;
+    let value = scaled_x + offset_y - scaled_time;
+    let clamped_value = (value % total_length + total_length) % total_length;
+    let alpha = step(material.spacing, clamped_value);
```
Indy2222, line 25: Making it semi-transparent might look even nicer.

Indy2222 (changes requested):
It is very nice looking. Unfortunately there is a bug in the implementation whose fix IMO needs a bit different approach.

Commit nitpick: I would put `Closes ...` into another line of commit description so that the first line stays under 50 characters (soft limit).

Indy2222:
To be create follow up tickets once merged:

- [ ] Update #2 so that it mentions that lines need to be updated (so that they are not below / above terrain).
- [ ] Do not draw the lines if they are too far away to avoid clutter.

assets/shaders/rally_point.wgsl
```diff
@@ -0,0 +1,27 @@
...
+
+@fragment
+fn fragment(
+    #import bevy_pbr::mesh_vertex_output
+) -> @location(0) vec4<f32> {
+    let scaled_x: f32 = uv.x * length(mesh.model[0].xyz);
+    let offset_y: f32 = abs(uv.y - 0.5) * material.pointiness;
+    let scaled_time: f32 = globals.time * material.speed;
+    let total_length = material.length + material.spacing;
+    let value = scaled_x + offset_y - scaled_time;
```
Indy2222, line 23: I would make it transparent if `(scaled_x + offset_y + material.spacing) > length(mesh.model[0].xyz)` so that the arrows disappear once the tip reaches the end (instead of it being clamped).

Issue #479, opened 2023-06-08
Add a configuration option for inverted mouse wheel zoom


Zakru:
Currently the mouse wheel zooms in when scrolled down and out when scrolled up. At least in my experience, most software uses the opposite controls, so I think there should at least be a configuration option to change that.

JackCrumpLeys:
Have a look at the `Camera` struct in the conf crate. We need a bool probably and a default of false. Macros handle the de-serialising (read pr  [#453](https://github.com/DigitalExtinction/Game/pull/453)) so should be easy. 

Then have a look at how the camera crate reads config (you will want to make a get method). I will do this later in the weekend if you don't want to.

PR #487, opened 2023-06-11
Add a configuration option for inverted mouse wheel zoom

JackCrumpLeys:
#479

### DIFF
--- a/crates/camera/src/camera.rs
+++ b/crates/camera/src/camera.rs
@@ -274,7 +274,12 @@ impl DesiredDistance {
     }
 
     fn zoom_clamped(&mut self, conf: &CameraConf, factor: f32) {
-        self.0 = (self.0 * factor).clamp(conf.min_distance(), conf.max_distance());
+        self.0 = (if conf.scroll_inverted() {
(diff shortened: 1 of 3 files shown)


crates/camera/src/camera.rs
```diff
@@ -274,7 +274,12 @@ impl DesiredDistance {
     }
 
     fn zoom_clamped(&mut self, conf: &CameraConf, factor: f32) {
-        self.0 = (self.0 * factor).clamp(conf.min_distance(), conf.max_distance());
+        self.0 = (if conf.scroll_inverted() {
```
JackCrumpLeys, line 277: Is this the best place for the logic?

Indy2222: I think so.

Indy2222 (changes requested):
Please update the documentation as part of the implementation commits. Otherwise LGTM.

Indy2222 (commented):
Please squash the two commits and add `Closes #479.` to commit description. Otherwise LGTM.

Indy2222 (approved):

PR #489, opened 2023-06-12
Display arrows from factory spawn to target

### LINKED ISSUE #434: Display "arrows" from selected factories to delivery location

0HyperCube:
Closes #434.

First draft implemented in PR #484.

Follow ups:

- Update #2: mention that the lines must be updated to work properly on non-flat surface.
- Add functionality which hides the line when the camera is too far away to avoid clutter.

### DIFF
--- a/assets/shaders/rally_point.wgsl
+++ b/assets/shaders/rally_point.wgsl
@@ -1,0 +1,36 @@
+#import bevy_pbr::mesh_view_bindings
+#import bevy_pbr::mesh_bindings
+
+struct CustomMaterial {
+    color: vec4<f32>,
+    pointiness: f32,
+    speed: f32,
+    length: f32,
+    spacing: f32,
+    fade: f32,
+};
+
+@group(1) @binding(0)
+var<uniform> material: CustomMaterial;
+
+@fragment
+fn fragment(
+    #import bevy_pbr::mesh_vertex_output
+) -> @location(0) vec4<f32> {
+    let world_space_length: f32 = length(mesh.model[0].xyz);
+    let scaled_x: f32 = uv.x * world_space_length;
+    let offset_y: f32 = abs(uv.y - 0.5) * material.pointiness;
+    let scaled_time: f32 = globals.time * material.speed;
+    let total_length = material.length + material.spacing;
+
+    let value = scaled_x + offset_y - scaled_time;
+    // Ensure that the result of the modulo operation is always positive
+    let positive_modulo = (value % total_length + total_length) % total_length;
+    let alpha = step(material.spacing, positive_modulo);
+
+    let start_fade: f32 = (floor(value / total_length) * total_length + scaled_time) / material.fade;
+    let end_fade: f32 = (world_space_length - ((ceil(value / total_length) * total_length + scaled_time))) / material.fade;
+    let fade = min(1., min(start_fade, end_fade));
+
+    return material.color * vec4(1., 1., 1., alpha * fade);
+}
--- a/assets/shaders/rally_point.wgsl
+++ b/assets/shaders/rally_point.wgsl
@@ -13,24 +13,31 @@ struct CustomMaterial {
 @group(1) @binding(0)
 var<uniform> material: CustomMaterial;
 
+const COLOR: vec4<f32> = vec4<f32>(0.0, 0.5, 0.0, 0.8);
+const POINTINESS: f32 = 2.;
+const SPEED: f32 = 3.;
+const LENGTH: f32 = 1.;
+const SPACING: f32 = 0.5;
+const FADE: f32 = 3.;
+
 @fragment
 fn fragment(
     #import bevy_pbr::mesh_vertex_output
 ) -> @location(0) vec4<f32> {
     let world_space_length: f32 = length(mesh.model[0].xyz);
     let scaled_x: f32 = uv.x * world_space_length;
-    let offset_y: f32 = abs(uv.y - 0.5) * material.pointiness;
-    let scaled_time: f32 = globals.time * material.speed;
-    let total_length = material.length + material.spacing;
+    let offset_y: f32 = abs(uv.y - 0.5) * POINTINESS;
+    let scaled_time: f32 = globals.time * SPEED;
+    let total_length = LENGTH + SPACING;
 
     let value = scaled_x + offset_y - scaled_time;
     // Ensure that the result of the modulo operation is always positive
     let positive_modulo = (value % total_length + total_length) % total_length;
-    let alpha = step(material.spacing, positive_modulo);
+    let alpha = step(SPACING, positive_modulo);
 
-    let start_fade: f32 = (floor(value / total_length) * total_length + scaled_time) / material.fade;
-    let end_fade: f32 = (world_space_length - ((ceil(value / total_length) * total_length + scaled_time))) / material.fade;
+    let start_fade: f32 = (floor(value / total_length) * total_length + scaled_time) / FADE;
+    let end_fade: f32 = (world_space_length - ((ceil(value / total_length) * total_length + scaled_time))) / FADE;
     let fade = min(1., min(start_fade, end_fade));
 
-    return material.color * vec4(1., 1., 1., alpha * fade);
+    return COLOR * vec4(1., 1., 1., alpha * fade);
 }
--- a/crates/construction/src/manufacturing.rs
+++ b/crates/construction/src/manufacturing.rs
@@ -15,7 +15,7 @@ use de_core::{
 use de_index::SpatialQuery;
 use de_objects::SolidObjects;
 use de_pathing::{PathQueryProps, PathTarget, UpdateEntityPath};
-use de_signs::UpdatePoleLocationEvent;
+use de_signs::{UpdateLineEndEvent, UpdateLineLocationEvent, UpdatePoleLocationEvent};
 use de_spawner::{ObjectCounter, SpawnBundle};
 use parry2d::bounding_volume::Aabb;
 use parry3d::math::Isometry;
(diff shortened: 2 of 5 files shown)


assets/shaders/rally_point.wgsl
```diff
@@ -0,0 +1,35 @@
...
+    #import bevy_pbr::mesh_vertex_output
+) -> @location(0) vec4<f32> {
+    let world_space_length: f32 = length(mesh.model[0].xyz);
+    let scaled_x: f32 = uv.x * world_space_length;
+    let offset_y: f32 = abs(uv.y - 0.5) * material.pointiness;
+    let scaled_time: f32 = globals.time * material.speed;
+    let total_length = material.length + material.spacing;
+
+    let value = scaled_x + offset_y - scaled_time;
+    let clamped_value = (value % total_length + total_length) % total_length;
```
Indy2222, line 27: What's the difference between `(value % total_length + total_length) % total_length;` and `value % total_length;`?

crates/construction/src/manufacturing.rs
```diff
@@ -327,6 +331,9 @@ fn change_locations(
     mut events: EventReader<ChangeDeliveryLocationEvent>,
     mut locations: Query<&mut DeliveryLocation>,
```
Indy2222, line 332: Every single factory (with `DeliveryLocation`) must have a `Transform` and `ObjectType`. I would therefore combine `transforms` and `locations` to a single query.

crates/construction/src/manufacturing.rs
```diff
@@ -335,6 +342,16 @@ fn change_locations(
                 event.factory(),
                 event.position(),
             ));
+            let Ok((transform, object_type)) = transforms.get(event.factory()) else {
+                continue;
+            };
+            let solid = solids.get(*object_type);
+            let Some(factory) = solid.factory() else {
```
Indy2222, line 349: I would unwrap here since the component `DeliveryLocation` implies that it is a factory.

crates/signs/src/line.rs
```diff
@@ -0,0 +1,209 @@
...
+    mut lines: ResMut<LineEntities>,
+    mut events: EventReader<SpawnLineEvent>,
+    mut commands: Commands,
+    mut meshes: ResMut<Assets<Mesh>>,
+    mut materials: ResMut<Assets<LineMaterial>>,
+) {
+    for SpawnLineEvent(transform, owner) in &mut events {
+        let line_id = commands
+            .spawn(MaterialMeshBundle {
+                mesh: meshes.add(shape::Plane::from_size(1.0).into()),
```
Indy2222, line 161: Performance-wise, it would be better to add the mesh only once (during `setup()`) and reuse it. That way, it would be stored only once in the VRAM and instancing would be used.

Indy2222 (changes requested):
We are half way there, but the PR still requires some care.

assets/shaders/rally_point.wgsl
0HyperCube: In wgsl `(-3) % 7` gives -3, when I wanted it to be +4.

Indy2222: I see, I didn't realize that. Could you add a code comment explaining that?

crates/signs/src/line.rs
```diff
@@ -0,0 +1,209 @@
...
+) {
+    for event in &mut events {
+        let line_entity = lines.0.entry(event.owner);
+        if event.visible && matches!(line_entity, Entry::Vacant(_)) {
+            if let Some(transform) = line_transforms.0.get(&event.owner) {
+                spawn_line_events.send(SpawnLineEvent(*transform, event.owner));
+            }
+        } else if !event.visible {
+            if let Entry::Occupied(entry) = line_entity {
+                commands.entity(entry.remove()).despawn_recursive();
```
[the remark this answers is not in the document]
0HyperCube: I already keep track of the start / end positions, I just accidentally made setting the transform also make the line visible.

crates/signs/src/line.rs
```diff
@@ -0,0 +1,209 @@
...
+            }
+        }
+    }
+}
+
+fn spawn_line(
```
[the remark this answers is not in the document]
0HyperCube: I have added a `despawn_line` system, however I'm unsure how to ensure that it is called when the factory is destroyed?

crates/signs/src/line.rs
```diff
@@ -0,0 +1,209 @@
...
+    fn alpha_mode(&self) -> AlphaMode {
+        self.alpha_mode
+    }
+}
+
+// Passed to the `rally_point.wgsl` shader
+#[derive(AsBindGroup, TypeUuid, Debug, Clone)]
+#[uuid = "d0fae52d-f398-4416-9b72-9039093a6c34"]
+pub struct LineMaterial {
+    #[uniform(0)]
```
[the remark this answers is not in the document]
0HyperCube: I get this error when I try and define a constant for a colour:
```
ERROR bevy_render::render_resource::pipeline_cache: failed to create shader module: Validation Error

Caused by:
    In Device::create_shader_module
    
Shader '' parsing error: this constant expression is not supported
    ┌─ wgsl:187:15
    │
187 │ const color = vec4(0.0, 0.5, 0.0, 0.8);
    │               ^^^^^^^^^^^^^^^^^^^^^^^^ expression is not supported
    │
    = note: this should be fixed in a future version of Naga
    = note: https://github.com/gfx-rs/naga/issues/1829


    this constant expression is not supported
```

crates/signs/src/line.rs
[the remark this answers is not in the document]
Indy2222: You can iterate over [RemovedComponents<AssemblyLine>](https://docs.rs/bevy/latest/bevy/ecs/prelude/struct.RemovedComponents.html) in the system to detect newly despawned factories.

crates/signs/src/line.rs
```diff
@@ -0,0 +1,268 @@
...
+    LineEnd,
+    LocationEvents,
+    VisibilityEvents,
+    SpawnLines,
+}
+
+#[derive(Resource)]
+struct LineMesh(Handle<Mesh>);
+
+fn setup(mut commands: Commands, mut meshes: ResMut<Assets<Mesh>>) {
```
Indy2222, line 89: Style: I usually put all structs at the top and all systems at the bottom of the module.

crates/signs/src/line.rs
```diff
@@ -0,0 +1,268 @@
...
+
+fn update_line_visibility(
+    mut events: EventReader<UpdateLineVisibilityEvent>,
+    mut lines: ResMut<LineEntities>,
+    line_transforms: Res<LineTransforms>,
+    mut spawn_line_events: EventWriter<SpawnLineEvent>,
+    mut despawn_line_events: EventWriter<DespawnLineEvent>,
+) {
+    for event in &mut events {
+        let line_entity = lines.0.entry(event.owner);
```
Indy2222, line 185: Note that multiple events concerning single entity might be received during a game tick. This might lead to duplicate lines, etc. Some of the incorrect behaviour might not be possible due to how e.g. selection systems work currently, but to minimize assumptions about other systems / event sent, we should modify the implementation. I suggest the following:

All needed structs:

* `LineLocations` mapping owner `Entity` to a struct with `start`, `end`, `transform` fields.
* `LineEntities` mapping owner `Entity` to line `Entity`.

All needed systems:

* **setup**
* **cleanup**
* **update_line_end** - loads `start` from `LineLocations` and sends event to `update_line_location`
* **update_line_location** - Is executed after `update_line_end`. It updates `LineLocations` (with "upserts") by changing start/end and recomputing `transform`
* **update_line_visibility** - is executed after `update_line_location`. It spawns / despawns line entities based on `visibility`. It is conditioned on (non)existence of the owner in `LineEntities` (because multiple events of the same visibility might be sent in a row). It logs a warning if line is to be spawned for owner not present in `LineLocations`. It uses `transform` from `LineLocations` when spawning.
* **owner_despawn** - iterates over despawned factories. For each such factory it despawns line entity and removes the line from both `LineLocations` and `LineEntities`. It should run after `update_line_location` and `update_line_visibility`.

crates/signs/src/line.rs
```diff
@@ -0,0 +1,268 @@
...
+    mut lines: ResMut<LineEntities>,
+) {
+    for event in &mut events {
+        if let Some(line_entity) = lines.0.remove(&event.owner) {
+            commands.entity(line_entity).despawn_recursive();
+        }
+    }
+}
+
+/// A transform matrix from a plane with points at `(-0.5, 0. -0.5),(0.5, 0. -0.5),(0.5, 0. 0.5),(-0.5, 0.-0.5)` to the line start and end with the desired width
```
Indy2222, line 230: This line looks excessively long, can we break comments at 79 chars?

crates/signs/src/line.rs
```diff
@@ -0,0 +1,268 @@
...
+fn spawn_line(
+    mut lines: ResMut<LineEntities>,
+    mut events: EventReader<SpawnLineEvent>,
+    mut commands: Commands,
+    line_mesh: Res<LineMesh>,
+    mut materials: ResMut<Assets<LineMaterial>>,
+) {
+    for SpawnLineEvent(transform, owner) in &mut events {
+        let line_id = commands
+            .spawn(MaterialMeshBundle {
```
Indy2222, line 207: We should add `DespawnOnGameExit` component.

Indy2222 (changes requested):

Indy2222 (approved):

Indy2222:
Udeps pipeline is failing due to https://github.com/rust-lang/rust/issues/112831

PR #521, opened 2023-06-18
Introduce audio config with music volume

### LINKED ISSUE #519: Make music volume configurable

JackCrumpLeys:
Closes #519.


crates/audio/Cargo.toml
```diff
@@ -14,6 +14,7 @@ categories.workspace = true
 [dependencies]
 # DE
 de_core.workspace = true
+de_conf.workspace = true
```
Indy2222, line 17: Please sort dependencies alphabetically.

crates/conf/src/conf.rs
```diff
@@ -45,6 +45,15 @@ pub struct Camera {
     #[ensure(*rotation_sensitivity > 0., "`rotation_sensitivity` must be greater than 0.0.")]
     pub rotation_sensitivity: f32,
 }
+
+#[derive(Deserialize, Config, Debug, Clone)]
+pub struct AudioConf {
+    pub music_enabled: bool,
+
+    #[is_finite]
+    #[ensure(*music_volume >= 0., "`music_volume` must be greater than 0.0.")]
```
Indy2222, line 54: * I would enforce that it is strictly larger than `0.0`. Alternatively, get rid of `music_enabled` and interpret `0.0` as disabled.
* Ensure that it is `<= 1.`.

crates/conf/src/conf.rs
```diff
@@ -154,8 +172,21 @@ impl MultiplayerConf {
     }
 }
 
+impl AudioConf {
+    /// Whether music is enabled.
```
Indy2222, line 176: Please remove this doc comments, they just repeat name of the methods.

Indy2222 (changes requested):
I had only a few minor comments on the code.

One comment **unrelated** to this PR: Config properties are `pub` and thus can not just be accessed but also changed from anywhere in the code.

I think that properties should be accessible only via getter methods so it is not possible to change them. Once configuration changing is to be implemented, it should be done via newly introduced setters, because:

* we could make it more defensive and assert that the new value is valid,
* and we could add any custom code tracking what was changed or cascading / propagating changes to related values if necessary.

Indy2222:
I forgot to mention it the last time: please update the docs/, there is a page dedicated to the configuration.

Indy2222:
I created #537 so such changes get easier in the future.

Indy2222:
Sorry, this has happen in the meantime #542. Should be easy to fix.

JackCrumpLeys:
> Sorry, this has happen in the meantime #542. Should be easy to fix.

fixed now

docs/src/conf.md
```diff
@@ -44,3 +44,6 @@ est as if all properties are missing.
     rotation_sensitivity` radians. It must be a positive finite number.
   * `scroll_inverted` (bool; default: `false`) – if `true`, mouse wheel and
     touchpad scrolling is inverted.
+* `audio` (object) – audio configuration.
+  * `music_volume` (f32; default: `1.0`) – sets the music volume. It must be a finite
+    number between `0.0` and `1.0`. If set to 0 music will not play.
```
Indy2222, line 49: Please add a new line at the end of file.

Indy2222 (changes requested):

Indy2222 (approved):

Indy2222 (approved):

Indy2222 (approved):

Issue #475, opened 2023-06-07
Energy Distrubution Grid: Power Flow


Indy2222:
Based on #472 and #473, implement a simple (v1) algorithm of energy distribution (power flow).

- Power may flow from any source (any battery) to any target.
  - Even from a unit to a power plan if needed.
- No energy interlink (an edge in the grid graph) can transmit than X watts (a constant in the code).
- Power may not flow from batteries with 0 charge or to batteries with maximum charge.
- The flow is "optimal" in some sense.
  - This must be designed so that "the optimal":
    - makes sense in the context of the game
    - can be computed with a non-super-complex algorithm
    - can be computed efficiently
  - For example: flow to batteries/units close to full discharge should be prioritized

Note that the goal of the issue is a version 1 implementation / algorithm. Simplicity is of higher priority than perfection.

Issue #474, opened 2023-06-07
Power Plants


Indy2222:
- [ ] There is a new building type called "Power Plant".
- [ ] It generates energy at a constant rate (a constant in the code).
- [ ] It consumes energy like any other building (see #473).
- [ ] The generated energy goes into its battery until maximum battery charge is reached.

Note that energy distribution is not implemented as part of this issue.

PR #522, opened 2023-06-18
Require energy in battery for movement

JackCrumpLeys:
#473 
As I am unfamiliar with the code in this crate this is mostly a best guess of where to put the logic. Units be able should keep paths and resume them when battery is charged. I have not tested they can resume but I am pretty sure they can; no easy way to add energy yet.

### DIFF
--- a/crates/movement/src/pathing.rs
+++ b/crates/movement/src/pathing.rs
@@ -67,7 +67,7 @@ fn check_battery(mut objects: Query<(&mut DesiredVelocity<PathVelocity>, &Batter
     for (mut movement, battery) in objects.iter_mut() {
         if battery.energy() <= 0. {
             movement.pause();
-        } else {
+        } else if movement.paused() {
             movement.resume();
         }
     }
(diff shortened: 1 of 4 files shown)


crates/movement/src/pathing.rs
```diff
@@ -67,6 +83,9 @@ fn follow_path(
     objects
         .par_iter_mut()
         .for_each_mut(|(transform, mut path, mut movement)| {
+            if movement.paused() {
```
Indy2222, line 86: Instead of having a separate system `check_battery()` I would change this condition to:

```rust
if battery.energy() <= 0. {
    movement.stop();
    return;
}
```

Indy2222 (changes requested):
This is blocked by #474 and #475. If we merge this before the two tickets are resolved, it would make the game unplayable (all units would sooner or later stop moving).

crates/movement/src/pathing.rs
```diff
@@ -18,6 +19,11 @@ impl Plugin for PathingPlugin {
                 .in_base_set(GameSet::PreMovement)
                 .run_if(in_state(GameState::Playing)),
         )
+        .add_system(
+            check_battery
```
Indy2222, line 23: `follow_path` should go `.after(PathingSet::CheckBattery)`.

Indy2222 (changes requested):
One last comment.

We should keep this open until battery charging is implemented.

PR #545, opened 2023-06-24
Add extra logs for config loading

JackCrumpLeys:


crates/conf/src/macros.rs
```diff
@@ -91,6 +92,7 @@ macro_rules! bundle_config {
 
             pub async fn load(path: &Path) ->  Result<Self, ConfigLoadError> {
                 let from = RawConfiguration::load(path).await?;
+                debug!("Loaded raw configuration: \n{}", serde_yaml::to_string(&from).expect("Failed to serialize raw configuration"));
```
Indy2222, line 95: This line is too long, maybe?:

```rust
                let serialized = serde_yaml::to_string(&from)
                    .expect("Failed to serialize raw configuration");
                debug!("Loaded raw configuration: \n{serialized}");
```

and whatever else rustfmt does with it.

crates/conf/src/macros.rs
```diff
@@ -75,7 +76,7 @@ macro_rules! bundle_config {
             }
         }
 
-        #[derive(Deserialize, Debug, Clone, Default)]
+        #[derive(MacroDeserialize, Serialize, Debug, Clone, Default)]
```
Indy2222, line 79: Shouldn't `Serialize` be `MacroSerialize`?

Indy2222 (changes requested):

Indy2222 (approved):

PR #557, opened 2023-06-30
Update dependencies

### LINKED ISSUE #556: CI is failing due to cargo udeps
This command `cargo udeps` fails with the following error:

```
error[E0635]: unknown feature `proc_macro_span_shrink`
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/proc-macro2-1.0.59/src/lib.rs:92:30
   |
92 |     feature(proc_macro_span, proc_macro_span_shrink)
   |                              ^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0635`.
error: could not compile `proc-macro2` (lib) due to previous error
```

JackCrumpLeys:
Fixes #556 and updates some other dependencies.


Cargo.toml
```diff
@@ -99,7 +99,7 @@ de_uom = { path = "crates/uom", version = "0.1.0-dev" }
 
 # Other
 ab_glyph = "0.2.21"
-ahash = "0.7.6"
+ahash = "0.8.3"
```
Indy2222, line 102: `bevy_utils` requires version `^0.7.0` thus I would not upgrade this dependency until Bevy upgrade.

Cargo.toml
```diff
@@ -110,18 +110,18 @@ bevy = { version = "0.10", features = ["mp3"] }
 bincode = "2.0.0-rc.3"
 chrono = "0.4.24"
 clap = { version = "4.0", features = ["derive"] }
-criterion = "0.4"
+criterion = "0.5.1"
 dirs = "5.0.0"
 enum-iterator = "1.4.0"
 enum-map = "2.3.0"
-fastrand = "1.9.0"
```
Indy2222, line 117: `futures-lite` require `^1.3.4` so I would not change this dependency until `futures-lite` upgrade.

Indy2222 (changes requested):
We should try to minimize the required number of versions of each dependency unless when we do not have a necessity to break this rule. Therefore, I would avoid several of the upgrades.

Issue #249, opened 2022-11-11
Make it possible to select enemy objects (in debug build)


Indy2222:
Introduce a build argument which makes it possible to select and control enemy units. This will make it easier to test some features.

Indy2222:
PR #558 introduced `godmode` feature flag. The flag allows selection of enemy units.

There is still some remaining work to do:

- [ ] Document the build flag in README
- [ ] Make it possible to attack local player units
- [ ] Make it impossible to attack self

PR #558, opened 2023-07-01
Introduce godmode feature flag

JackCrumpLeys:
This change makes it possible to select enemy units. Further modifications are still needed, for example to allow attacking local player.

Relates to #249.

### DIFF
--- a/crates/spawner/src/spawner.rs
+++ b/crates/spawner/src/spawner.rs
@@ -78,7 +78,7 @@ fn spawn(
                 let player = *player.expect("Active object without an associated was spawned.");
                 counter.player_mut(player).unwrap().update(active_type, 1);
 
-                if player == game_config.player() {
+                if player == game_config.player() || cfg!(debug_assertions){
                     entity_commands.insert(Playable);
                 }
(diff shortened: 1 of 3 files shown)


crates/spawner/src/spawner.rs
```diff
@@ -78,7 +78,7 @@ fn spawn(
                 let player = *player.expect("Active object without an associated was spawned.");
                 counter.player_mut(player).unwrap().update(active_type, 1);
 
-                if player == game_config.player() {
+                if player == game_config.player() || cfg!(debug_assertions){
```
Indy2222, line 81: IMHO `debug_assertions` should produce the same functionality (just a bit slower & with more assertions). Its main purpose is to make bug / wrong assumption discovery easier.

Sorry for not making #249 clear. We should hide this behind a [feature](https://doc.rust-lang.org/cargo/reference/features.html) so one can `cargo run --features enemy`.

Indy2222 (changes requested):

Indy2222 (approved):

PR #587, opened 2023-07-05
Feature/god mode fix

Indy2222:
Relates to #249.


Indy2222:
It will be a bit harder to fix, especially in combination with multiplayer. Not a priority for me right now.

PR #606, opened 2023-07-07
Rewrite of despawn logic to allow for despawn events with data.

JackCrumpLeys:
By adding a plugin for the component you wish to monitor can listen to component de-spawn events and get data. Components must also implement clone to work with this system.

```rs
#[derive(Clone, Component)]
struct Bar(f32);

let mut app = App::new();

// watch for despawning of entities with `Bar` component
app.add_plugins(DespawnEventsPlugin::<&Bar, Bar>::default())
app.add_systems(Update, foo.after(DespawnerSet::Despawn));

fn foo(despawned: EventReader<DespawnedComponentsEvent<Bar>>) {
    for event in despawned { // for every newly despawned `Foo` component,
        println("Dead number: {}", event.data.0); // print their number
    }
}
```

Have a look at the test (`de_spawner::destroyer::tests::despawn_event`) for a more in depth-example.


crates/spawner/src/destroyer.rs
```diff
@@ -9,15 +9,51 @@ pub(crate) struct DestroyerPlugin;
...
+pub trait DespawnEventProducer {
+    type Data;
+
+    fn data(&self) -> Self::Data;
+}
+
+// Contains the entity and the data of the despawned component
+pub struct DespawnedComponents<T: DespawnEventProducer + Send + Sync>
+where
+    <T as DespawnEventProducer>::Data: Send + Sync,
```
Indy2222, line 46: I would impose these trait bounds in `DespawnEventProducer` (`type Data: Send + Sync;`) because it does not make sense to have a `DespawnEventProducer` without the data being Send + Sync.

crates/spawner/src/destroyer.rs
```diff
@@ -27,7 +63,28 @@ fn destroy(
             if let ObjectType::Active(active_type) = object_type {
                 counter.player_mut(player).unwrap().update(active_type, -1);
             }
-            commands.entity(entity).despawn_recursive();
+            commands.entity(entity).insert(ToBeDespawned);
```
Indy2222, line 66: I won't do this via marker trait `ToBeDespawned` but via an event to avoid the need to wait until the next base (for commands being flushed to World).

What we can do:

* keep `despawn_recursive()` here (the entity stays available until next stage)
* send `Despavning(Entity)` event from here
* register `send_data<T>` via a public plugin `DespawnEventsPlugin<T>` so that it runs in Update stage (while the entities are still available) but `.after()` `find_dead`
* From everywhere it is needed, one can add the appropriate `DespawnEventsPlugin<T>`

What do you think?

Indy2222 (changes requested):

crates/spawner/src/destroyer.rs
```diff
@@ -9,15 +9,51 @@ pub(crate) struct DestroyerPlugin;
...
-fn destroy(
+#[derive(Debug, Clone, PartialEq, Eq, Hash, SystemSet)]
+enum DestroyerSet {
+    FindDead,
+    SendData,
+    DestroyDead,
+}
+
+/// Special marker trait which when used on a component the despawner sends an event with the data
+pub trait DespawnEventProducer {
```
Indy2222, line 37: Thinking about it more: we can introduce `DespawnEventsPlugin<T: ReadOnlyWorldQuery>` ([ReadOnlyWorldQuery](https://docs.rs/bevy/latest/bevy/ecs/query/trait.ReadOnlyWorldQuery.html#)) which would automatically send events `DespawnedEvent<<T as WorldQuery>::Item>`. We might need to limit Item to be Copy so we can easily produce the events.

This would make it much more versatile:\

* one would be able to query any combination of components
* the events could be processed during whole frame (even after Update base set when the entity is already despawned)
* there would not be any trait one has to implement, all needed would be to add `DespawnEventsPlugin<T>`

crates/spawner/src/lib.rs
```diff
@@ -24,7 +24,6 @@ impl PluginGroup for SpawnerPluginGroup {
             .add(CounterPlugin)
             .add(SpawnerPlugin)
             .add(DraftPlugin)
-            .add(DestroyerPlugin)
```
Indy2222, line 27: This is just some development remnant, right?

crates/spawner/src/destroyer.rs
```diff
@@ -1,33 +1,254 @@
...
+
+struct Despawning(Entity);
+
+#[allow(unused)]
+fn send_data<'w, Q, T, F>(
+    mut despawning: EventReader<Despawning>,
+    mut events: EventWriter<DespawnedComponents<Q, T, F>>,
+    data: Query<Q, F>,
+) where
+    T: Clone + Component + Send + Sync + 'w,
```
Indy2222, line 126: I think that this should be changed to something like

```rust
fn send_data<'w, Q, F, T>(
    mut despawning: EventReader<DespawnEvent>,
    mut events: EventWriter<DespawnedEvent<<<<Q as WorldQuery>::ReadOnly as WorldQuery>::Item<'w> as ToOwned>::Owned>>,
    data: Query<Q, F>,
) where
    Q: WorldQuery + Send + Sync + 'w,
    F: ReadOnlyWorldQuery + Send + Sync + 'w,
    <<Q as WorldQuery>::ReadOnly as WorldQuery>::Item<'w>: Clone + Send + Sync,
```

Indy2222 (changes requested):
I would merge only changes to `spawner` crate. The rest is either not needed or can be done in separate PRs.

crates/spawner/src/destroyer.rs
```diff
@@ -1,33 +1,254 @@
...
+                .in_set(SpawnerSet::Destroyer)
+                .in_set(DestroyerSet::DestroyDead),
         );
     }
 }
 
-fn destroy(
+#[derive(Debug, Clone, PartialEq, Eq, Hash, SystemSet)]
+enum DestroyerSet {
+    FindDead,
```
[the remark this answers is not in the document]
JackCrumpLeys: I think we need a `PreDespawn` for other plugins to use if they expect that units might die/if they are mutating health.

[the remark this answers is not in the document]
Indy2222: Agreed, although a better name would be nice. What about `Destruction`?

crates/spawner/src/destroyer.rs
```diff
@@ -1,33 +1,254 @@
...
+                .in_set(DestroyerSet::SendData),
+        );
+    }
+}
+
+// Contains the entity and the data of the despawned component
+#[derive(Debug)]
+pub struct DespawnedComponents<Q, T, F = ()>
+where
+    F: ReadOnlyWorldQuery + Send + Sync + 'static,
```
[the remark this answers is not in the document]
JackCrumpLeys: I think that for ergonomic reasons we should keep the filter. For example we could have two different system each wanting a different filtering of the same component. If we allow the event to be distinct we allow them to run at the same time. For example: `DespawnedComponents<Foo, With<Bar>>` vs `DespawnedComponents<Foo, With<Baz>`.

crates/spawner/src/destroyer.rs
JackCrumpLeys: I don't think this is possible. I gave it my best shot but I think the current implementation is as close as we will get. feel free to have a go on your own.

crates/spawner/src/destroyer.rs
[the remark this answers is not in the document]
Indy2222: You are right.

crates/spawner/src/destroyer.rs
```diff
@@ -1,33 +1,253 @@
...
-                .in_base_set(GameSet::Update)
-                .run_if(in_state(AppState::InGame))
-                .in_set(SpawnerSet::Destroyer),
-        );
+        app.add_event::<DespawnEvent>()
+            .add_system(
+                find_dead
+                    .in_base_set(GameSet::Update)
+                    .run_if(in_state(AppState::InGame))
+                    .before(DespawnerSet::Despawn),
```
Indy2222, line 19: Shouldn't this be in `.in_set(DespawnerSet::Destruction)`?

crates/combat/src/laser.rs
```diff
@@ -15,7 +15,7 @@ impl Plugin for LaserPlugin {
             fire.in_base_set(GameSet::Update)
                 .run_if(in_state(GameState::Playing))
                 .in_set(AttackingSet::Fire)
-                .before(SpawnerSet::Destroyer),
+                .before(DespawnerSet::Despawn),
```
Indy2222, line 18: Shouldn't this be before `Destruction`?

crates/spawner/src/destroyer.rs
```diff
@@ -1,33 +1,253 @@
...
+                .run_if(in_state(AppState::InGame))
+                .after(DespawnerSet::Destruction)
+                .in_set(DespawnerSet::Despawn),
+        );
+    }
+}
+
+// Contains the entity and the data of the despawned component
+#[derive(Debug)]
+pub struct DespawnedComponents<T, F = ()>
```
Indy2222, line 104: I would rename this to `DespawnedComponentsEvent` so it is in accordance with the style guide (see CONTRIBUTING.md).

crates/spawner/src/destroyer.rs
```diff
@@ -1,33 +1,253 @@
...
+            send_data::<Q, T, F>
+                .in_base_set(GameSet::Update)
+                .run_if(in_state(AppState::InGame))
+                .after(DespawnerSet::Destruction)
+                .in_set(DespawnerSet::Despawn),
+        );
+    }
+}
+
+// Contains the entity and the data of the despawned component
```
Indy2222, line 102: I would adapt this comment and change it to docs: ``/// This event is sent by [`DespawnEventsPlugin`] when a matching entity is being despawned.``

crates/spawner/src/destroyer.rs
```diff
@@ -1,33 +1,253 @@
...
+}
+
+fn despawn(mut commands: Commands, mut despawning: EventReader<DespawnEvent>) {
+    for entity in despawning.iter() {
+        commands.entity(entity.0).despawn_recursive();
+    }
+}
+
+#[derive(Debug)]
+pub struct DespawnEventsPlugin<Q, T, F = ()>
```
Indy2222, line 63: I would add docs, something like: ``/// Add this plugin to receive [`DespawnedComponentsEvent`] when matching entities are despawned.``

crates/spawner/src/lib.rs
```diff
@@ -24,12 +25,6 @@ impl PluginGroup for SpawnerPluginGroup {
...
```
Indy2222, line 27: Looks like `DespawnerPlugin` is never added.

crates/spawner/src/lib.rs
```diff
@@ -3,7 +3,8 @@
 use bevy::{app::PluginGroupBuilder, prelude::*};
 use counter::CounterPlugin;
 pub use counter::ObjectCounter;
-use destroyer::DestroyerPlugin;
+pub use destroyer::DespawnerSet;
```
Indy2222, line 6: Also `DespawnEventsPlugin` and `DespawnedComponents` should be reexported here.

Indy2222 (changes requested):
I looks good high-level, I've left several small comments.

crates/spawner/src/destroyer.rs
JackCrumpLeys: I think it would be better to do `.before(DespawnerSet::Despawn)` because if the health check system runs before systems mutating health (which should be running in this set) that could cause a frame of delay. 

Mabye we could do `.after(DespawnerSet::Destruction)` if that would be better?

crates/combat/src/laser.rs
JackCrumpLeys: Oops yes should be IN set.

Indy2222 (changes requested):

Indy2222:
The PR must be rebased on `main` because of the upgrade to Bevy v0.11.

JackCrumpLeys:
Rebased! ready for merge :)

Indy2222 (dismissed):

Indy2222:
I suggest to do last minor changes (see the patch below):

```patch
From dafb891c79b33e4f6d49043547e073a7c16243f9 Mon Sep 17 00:00:00 2001
From: Martin Indra <martin.indra@mgn.cz>
Date: Tue, 25 Jul 2023 22:48:57 +0200
Subject: [PATCH] Cosmetics

---
 crates/spawner/src/{destroyer.rs => despawner.rs} | 0
 crates/spawner/src/lib.rs                         | 9 +++------
 2 files changed, 3 insertions(+), 6 deletions(-)
 rename crates/spawner/src/{destroyer.rs => despawner.rs} (100%)

diff --git a/crates/spawner/src/destroyer.rs b/crates/spawner/src/despawner.rs
similarity index 100%
rename from crates/spawner/src/destroyer.rs
rename to crates/spawner/src/despawner.rs
diff --git a/crates/spawner/src/lib.rs b/crates/spawner/src/lib.rs
index 9a9debf..5025d1b 100644
--- a/crates/spawner/src/lib.rs
+++ b/crates/spawner/src/lib.rs
@@ -3,20 +3,17 @@
 use bevy::{app::PluginGroupBuilder, prelude::*};
 use counter::CounterPlugin;
 pub use counter::ObjectCounter;
-pub use destroyer::DespawnEvent;
-pub use destroyer::DespawnEventsPlugin;
-pub use destroyer::DespawnedComponentsEvent;
-pub use destroyer::DespawnerSet;
+pub use despawner::{DespawnEvent, DespawnEventsPlugin, DespawnedComponentsEvent, DespawnerSet};
 use draft::DraftPlugin;
 pub use draft::{DraftAllowed, DraftBundle};
 use gameend::GameEndPlugin;
 pub use spawner::SpawnBundle;
 use spawner::SpawnerPlugin;
 
-use crate::destroyer::DespawnerPlugin;
+use crate::despawner::DespawnerPlugin;
 
 mod counter;
-mod destroyer;
+mod despawner;
 mod draft;
 mod gameend;
 mod spawner;
-- 
2.41.0
```

Indy2222 (dismissed):

Indy2222 (approved):

Indy2222 (approved):

Issue #394, opened 2023-02-07
Add basic sound set


Indy2222:
- [ ] Laser fire
- [ ] Building & unit destruction / explosion
- [ ] ?

PR #634, opened 2023-07-21
Add spatial audio effects

Zakru:
I'm aware of an upcoming upgrade to Bevy 0.11 and if this PR is delayed, I'm willing to migrate it if needed.

Replace `bevy_audio` with `bevy_kira_audio`.

Sounds are unedited/derived from CC0 sounds from FreeSound.org.

Works towards #394, although many vital effects are still missing.

List of sounds:
* Building construction
* Unit construction
* Building destruction
* Unit destruction (currently same but with a lower volume)

Sounds that aren't included but I've seen listed include units firing and UI sounds.

Attenuation is not great, but perhaps fine for now. imo, the dream would be for sounds to be loudest when they are close to the focal area of the view, and become quieter when they leave the screen/go far away.


Zakru:
Just adding some justification for using Kira. Bevy's built-in audio engine is very restricted. As it's based on Rodio, it only allows simple physically-based sound with little control. Kira seems to be the most popular alternative sound backend for Bevy, and allows very fine control over many aspects of sounds, including total control over volume and panning, which I think comes in very handy in a game such as this.

Zakru:
Also idk why but I can't seem to get rid of `bevy_audio` from the dependency tree, so I had to use aliases for Kira's similarly named types.

Indy2222:
> I'm aware of an upcoming upgrade to Bevy 0.11 and if this PR is delayed, I'm willing to migrate it if needed.

I am in the middle of migration to Bevy 0.11 (#636). I hope to finish the migration during the weekend. Let's see what lands first.

assets/audio/sounds/construct.ogg
Indy2222, line 1: Please update .gitattributes so that the sounds are tracked with GIT LFS (`*.waw, *.ogg`).

assets/audio/sounds/destruction.wav
Indy2222, line 1: Why is this a .waw and other two sounds .ogg? I would convert to .ogg of a decent bitrate so that the files are smaller.

crates/construction/src/manufacturing.rs
```diff
@@ -306,6 +314,18 @@ impl ProductionItem {
     }
 }
 
+fn setup(mut commands: Commands, server: Res<AssetServer>) {
```
Indy2222, line 317: I would encapsulate as much sound logic as possible to `de_audio` crate. This should include loading the assets and playing the sound. The interface could be via events like this: 

```rust
audio.send(PlayAudioEvent::new(Sound::Manufacture, location));
```

crates/spawner/src/destroyer.rs
```diff
@@ -1,32 +1,87 @@
...
+                commands.spawn((
+                    TransformBundle::from_transform(*transform),
+                    match active_type {
+                        ActiveObjectType::Building(_) => SpatialSoundBundle {
+                            sound: sounds.building.clone(),
+                            ..Default::default()
+                        },
+                        ActiveObjectType::Unit(_) => SpatialSoundBundle {
+                            sound: sounds.unit.clone(),
+                            volume: Volume(0.4),
```
Indy2222, line 78: Cannot we duplicate the sound file and make it less loud instead of this?

crates/audio/src/spatial.rs
```diff
@@ -0,0 +1,125 @@
...
+    focus: &CameraFocus,
+    sound: &GlobalTransform,
+) -> (f64, f64) {
+    let cam_right = camera.right();
+    let sound_dir = (sound.translation() - camera.translation()).normalize();
+    let pan = cam_right.dot(sound_dir) * 0.5 + 0.5;
+
+    let distance_from_camera = camera.translation().distance(sound.translation());
+    let camera_zoom_distance = focus.distance().inner();
+    let distance_attenuation =
```
Indy2222, line 65: I would try to apply inverse-square law first and only make it linear if it doesn't sound good.

```rust
let camera_distance_squared = camera.translation().distance_squared(sound.translation());
// Anything closer than 70% of zoom distance is at full volume.
let base_distance_squared = (0.7 * focus.distance().inner()).powi(2);
// Inverse-square law
let distance_attenuation = (base_distance_squared / camera_distance_squared).min(1.);
```

Indy2222 (changes requested):

crates/spawner/src/destroyer.rs
Zakru: This was intended to be temporary, so actually I should actually just add a unique sound. Although you're correct, even a duplicated sound would be a better placeholder than this. 👍

crates/audio/src/spatial.rs
Zakru: This is actually something I'd like to really look deeper into perhaps at a later time. I *feel* (although haven't experimented much, just tried to figure out what I want) like standard inverse-square/other distance-based attenuation might not be the best fit here.

Take an example scenario, I'm looking at a building with the furthest zoom level at a 45 degree angle from the ground. I obviously want to hear the sounds emitted by that building. But if there's another building the same distance away, directly behind the camera, I don't think it feels natural to play sounds from it, at least at the same level.

Not that this current implementation does anything in favor of that, just want to bring it up. I thought maybe I could get away with the sound overhaul in one PR and then tune the precise parameters, but I can experiment here as well ofc.

Zakru: I unfortunately don't have any other games quickly on hand, but I did some experimenting in LoL since it uses an RTS-like camera (and it has its deepest roots in RTS). In that game, sound volume seems to depend on distance from the camera focus (the area where the sound is audible seems like a circular area when clicking the minimap), with a short falloff near the edge of the audible range.

With that in mind, I've updated the PR with a proposal for a similar falloff function. I think something like it is a great fit for an RTS, although in this state it is certainly not perfect.

Indy2222: > Take an example scenario, I'm looking at a building with the furthest zoom level at a 45 degree angle from the ground. I obviously want to hear the sounds emitted by that building. But if there's another building the same distance away, directly behind the camera, I don't think it feels natural to play sounds from it, at least at the same level.

I agree but this is IMHO orthogonal to the distance square law. Instead, we should either discard all sounds coming from outside the camera frustum or have another multiplicative attenuation factor based on direction. I would go for the first version to avoid over-engineering.

> Not that this current implementation does anything in favor of that, just want to bring it up. I thought maybe I could get away with the sound overhaul in one PR and then tune the precise parameters, but I can experiment here as well ofc.

:+1: Yep. I would not deal with the direction issue here.

The distance attenuation experiment should be easy (few lines of code) so I would try that. Lets do a few quick experiment, pick the best easy solution and create an issue for further experimentation / research on the topic.

> I unfortunately don't have any other games quickly on hand, but I did some experimenting in LoL since it uses an RTS-like camera (and it has its deepest roots in RTS). In that game, sound volume seems to depend on distance from the camera focus (the area where the sound is audible seems like a circular area when clicking the minimap), with a short falloff near the edge of the audible range.

Interesting. Can you push commented out versions so that I can try them all myself? :pray:

I was imagining that e.g. a drone flying high up (close to the camera) being louder might create a cool effect.

I think that a combination of the distance attenuation (from camera) + directional attenuation (angle from forward direction) might give nice results. I.e. close camera flybys would be loud while things near the center of the screen would be more dominant.

Indy2222: Also, in the future, we should consider something like Reinhard mapping so that we don't have a hard clip at value 1.

crates/audio/src/spatial.rs
```diff
@@ -0,0 +1,168 @@
...
+                LoadState::Loaded => 1,
+                LoadState::NotLoaded | LoadState::Loading => 0,
+                _ => panic!("Unexpected loading state."),
+            })
+            .sum(),
+        total: sounds
+            .0
+            .len()
+            .try_into()
+            .expect("Trying to load an ungodly number of sounds"),
```
Indy2222, line 94: LOL

crates/audio/src/spatial.rs
Zakru: I have added a few options for both distance attenuation and view occlusion, accessible with debug controls that work like two sets of radio buttons.

**Distance attenuation**
1 = No attenuation
2 = Linear from focus
3 = Inverse square
4 = Inverse square from focus

**View occlusion**
8 = None
9 = Don't play sounds outside view
0 = Volume is multiplied by 1...0 from 22.5...45 degrees from center of view, calculated from a point a bit behind the camera so that sounds just behind the camera are also heard

Current personal favorite combo is inverse square attenuation + angle occlusion (3 + 0).

crates/audio/src/spatial.rs
```diff
@@ -0,0 +1,292 @@
...
+            // Slope parameter of linear function (-1 unit in 25.5 degrees or pi/8)
+            const SLOPE: f32 = -1.0 / (PI / 8.);
+            // Constant parameter of linear function
+            const CONSTANT: f32 = 2.0;
+            (SLOPE * angle + CONSTANT).clamp(0.0, 1.0)
+        }
+        None => 1.,
+    };
+
+    let volume = attenuation_factor * occlusion_factor;
```
Indy2222, line 230: I would clamp this instead of the operands individually so that something e.g. very close but at a bad angle doesn't become unnecessarily silent.

Indy2222 (changes requested):
Wow, the keybindings made it super easy to test! :heart:

I would use Attenuation::InverseSquare + Occlusion::Angle. IMHO it is good enough for merge. We can improve that later.

Ideas for further improvements in follow up PRs / experiments:

- make the angle sensitivity dependent on camera frustum angles, on wide monitors the fall off slope should be steeper vertically,
- compute the the angle from camera location (not moved backward) and make the angle fall-off non-linear (almost constant close to 0 and steep just behind camera frustum)
- use some kind of dynamic range compression so that everything closer than a threshold is not at the same volume.

Zakru:
PR itself is complete -> de-draftified, can port to 0.11

crates/audio/src/spatial.rs
```diff
@@ -0,0 +1,182 @@
...
+    spatial_audios: Query<InitializedSound, With<SpatialSound>>,
+    camera: Query<&GlobalTransform, With<Camera>>,
+    focus: Res<CameraFocus>,
+    mut audio_instances: ResMut<Assets<AudioInstance>>,
+) {
+    let camera = camera.single();
+
+    for (entity, audio, transform) in &spatial_audios {
+        let Some(audio_instance) = audio_instances.get_mut(audio) else {
+            commands.entity(entity).despawn();
```
Indy2222, line 173: I would use `despanw_recursive()` so that there are no orphans in case we add any children in the future.

Indy2222 (changes requested):

Indy2222 (approved):

PR #642, opened 2023-07-24
Improve minimap camera controls

Zakru:
* Minimap no longer registers camera input when mouse is pressed down elsewhere and brought over the minimap
* Minimap camera responds to press instead of release
* Camera can now be dragged across the minimap in addition to clicking

I felt that at least with the current system, these new camera controls don't fit the existing event model, so I separated them to handle raw mouse input.

### DIFF
--- a/crates/controller/src/hud/minimap/interaction.rs
+++ b/crates/controller/src/hud/minimap/interaction.rs
@@ -1,7 +1,10 @@
 use std::fmt;
 
 use bevy::{
-    input::{mouse::MouseButtonInput, ButtonState},
+    input::{
+        mouse::{MouseButtonInput, MouseMotion},
+        ButtonState,
+    },
     prelude::*,
     window::PrimaryWindow,
 };
@@ -19,35 +22,45 @@ pub(super) struct InteractionPlugin;
 
 impl Plugin for InteractionPlugin {
     fn build(&self, app: &mut App) {
-        app.add_event::<MinimapClickEvent>().add_systems(
-            InputSchedule,
-            (
-                click_handler.in_set(InteractionSet::ClickHandler),
-                move_camera_system.after(InteractionSet::ClickHandler),
-                send_units_system
-                    .after(InteractionSet::ClickHandler)
-                    .before(CommandsSet::SendSelected),
-                delivery_location_system
-                    .after(InteractionSet::ClickHandler)
-                    .before(CommandsSet::DeliveryLocation),
-            )
-                .run_if(in_state(GameState::Playing)),
-        );
+        app.add_event::<MinimapPressEvent>()
+            .add_event::<MinimapDragEvent>()
+            .insert_resource(DraggingButtons(Vec::new()))
+            .add_systems(
+                InputSchedule,
+                (
+                    click_handler
+                        .in_set(InteractionSet::ClickHandler)
+                        .run_if(on_event::<MouseButtonInput>()),
+                    drag_handler
+                        .in_set(InteractionSet::DragHandler)
+                        .after(InteractionSet::ClickHandler)
+                        .run_if(on_event::<MouseMotion>()),
+                    move_camera_system.after(InteractionSet::DragHandler),
+                    send_units_system
+                        .after(InteractionSet::ClickHandler)
+                        .before(CommandsSet::SendSelected),
+                    delivery_location_system
+                        .after(InteractionSet::ClickHandler)
+                        .before(CommandsSet::DeliveryLocation),
+                )
+                    .run_if(in_state(GameState::Playing)),
+            );
     }
 }
 
 #[derive(Copy, Clone, Hash, Debug, PartialEq, Eq, SystemSet)]
 enum InteractionSet {
     ClickHandler,
+    DragHandler,
 }
 
 #[derive(Event)]
-struct MinimapClickEvent {
+struct MinimapPressEvent {
     button: MouseButton,
     position: Vec2,
 }
 
-impl MinimapClickEvent {
+impl MinimapPressEvent {
     fn new(button: MouseButton, position: Vec2) -> Self {
         Self { button, position }
     }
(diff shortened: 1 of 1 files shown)


crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -105,15 +119,62 @@ fn click_handler(
     }
 }
 
-fn move_camera_system(
```
Indy2222, line 108: I think that having a separate code for camera movement & rest of minimap interaction is not wise. For example unit sending with right clicks suffers the same issues.

I agree that moving the camera on mouse release even when mouse was pressed outside of the minimap is not ideal. Maybe we can fix this by renaming `MinimapClickEvent` to `MinimapPressEvent` and changing `click_handler()` so that it sends the event on press. That would fix the issue with minimal effort.

crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -105,15 +119,62 @@ fn click_handler(
...
+fn move_camera_drag_system(
+    window_query: Query<&Window, With<PrimaryWindow>>,
+    mut input_events: EventReader<MouseMotion>,
+    hud: HudNodes<With<MinimapNode>>,
+    bounds: Res<MapBounds>,
+    mouse_buttons: Res<Input<MouseButton>>,
+    dragging: Res<DraggingCamera>,
+    mut camera_events: EventWriter<MoveFocusEvent>,
+) {
+    input_events.clear();
```
Indy2222, line 165: Why is this here?

crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -105,15 +119,62 @@ fn click_handler(
...
+            dragging.0 = true;
+            let proportional = Vec2::new(relative.x, 1. - relative.y);
+            let event = MoveFocusEvent::new(bounds.rel_to_abs(proportional));
+
+            camera_events.send(event);
+        }
+    }
+}
+
+fn move_camera_drag_system(
```
Indy2222, line 156: - I think that dragging implementation is independent on the fix to mouse click / press / release behavior. Therefore, I would separate this changes to a standalone system.
- I would change this system to `dragg_handler()` (similar to `click_handler()`) sending `MinimapDraggEvent` and handle the events in `move_camera_system()`.

crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -105,15 +119,62 @@ fn click_handler(
...
+    window_query: Query<&Window, With<PrimaryWindow>>,
+    mut input_events: EventReader<MouseMotion>,
+    hud: HudNodes<With<MinimapNode>>,
+    bounds: Res<MapBounds>,
+    mouse_buttons: Res<Input<MouseButton>>,
+    dragging: Res<DraggingCamera>,
+    mut camera_events: EventWriter<MoveFocusEvent>,
+) {
+    input_events.clear();
+    if !mouse_buttons.pressed(MouseButton::Left) || !dragging.0 {
```
Indy2222, line 166: I think that we do not need `dragging` resource at all. Instead we can simplify this by having a `Local<Vec2>` storing the last mouse position and sending the drag event when the current mouse position is different to the last one while `mouse_buttons.pressed(MouseButton::Left)`.

Indy2222 (changes requested):

crates/controller/src/hud/minimap/interaction.rs
Zakru: Ah, I was somehow under the misconception that `on_event` requires the system to use the events 😳

crates/controller/src/hud/minimap/interaction.rs
Zakru: We need to be able to tell whether the click started inside the minimap, so that we don't react to a drag that was started outside of it. In fact, if we generalize the drag events as well, I believe we need to store a list/set of the held buttons to keep track of each one separately. What do you think?

crates/controller/src/hud/minimap/interaction.rs
Indy2222: No, the old events clear automatically.

crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -63,50 +76,117 @@ impl MinimapClickEvent {
...
+struct DraggingButtons(Vec<MouseButton>);
+
 fn click_handler(
     window_query: Query<&Window, With<PrimaryWindow>>,
     mut input_events: EventReader<MouseButtonInput>,
     hud: HudNodes<With<MinimapNode>>,
     bounds: Res<MapBounds>,
-    mut click_events: EventWriter<MinimapClickEvent>,
+    mut dragging: ResMut<DraggingButtons>,
+    mut click_events: EventWriter<MinimapPressEvent>,
```
Indy2222, line 116: I would finish with the renaming:

* `click_handler()` -> `press_handler()`
* `mut click_events` -> `mut press_events`
* `InteractionSet::ClickHandler` -> `InteractionSet::PressHandler`

crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -19,35 +22,45 @@ pub(super) struct InteractionPlugin;
...
+                InputSchedule,
+                (
+                    click_handler
+                        .in_set(InteractionSet::ClickHandler)
+                        .run_if(on_event::<MouseButtonInput>()),
+                    drag_handler
+                        .in_set(InteractionSet::DragHandler)
+                        .after(InteractionSet::ClickHandler)
+                        .run_if(on_event::<MouseMotion>()),
+                    move_camera_system.after(InteractionSet::DragHandler),
```
Indy2222, line 38: Even though proper ordering is implied, I would be more explicitly here and not depend on other system dependencies: `.after(InteractionSet::ClickHandler).after(InteractionSet::DragHandler)`.

crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -63,50 +76,117 @@ impl MinimapClickEvent {
...
+    mut click_events: EventWriter<MinimapPressEvent>,
 ) {
     let Some(cursor) = window_query.single().cursor_position() else {
         return;
     };
+
     for event in input_events.iter() {
-        if event.state != ButtonState::Released {
+        if event.state != ButtonState::Pressed {
+            dragging.0.retain(|b| *b != event.button);
```
Indy2222, line 124: This must be done before the early return (let-else above), otherwise the button would stay "pressed" indefinitely if released outside of the game window.

Also, please test that it works correctly when:

1. press the button while in the minmap
1. hold the button and switch to a different window (e.g. with Alt-TAB)
1. release the button
1. switch back to the game

crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -63,50 +76,117 @@ impl MinimapClickEvent {
...
+        }
+
+        let event = MoveFocusEvent::new(drag.position());
+        camera_events.send(event);
     }
 }
 
 fn send_units_system(
-    mut click_events: EventReader<MinimapClickEvent>,
+    mut click_events: EventReader<MinimapPressEvent>,
```
Indy2222, line 189: `click_events` -> `press_events`

crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -118,7 +198,7 @@ fn send_units_system(
 }
 
 fn delivery_location_system(
-    mut click_events: EventReader<MinimapClickEvent>,
+    mut click_events: EventReader<MinimapPressEvent>,
```
Indy2222, line 201: `press_events`

crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -63,50 +76,117 @@ impl MinimapClickEvent {
...
+        for b in &dragging.0 {
+            let event = MinimapDragEvent::new(*b, world);
+            drag_events.send(event);
+        }
+    }
+}
+
 fn move_camera_system(
-    mut click_events: EventReader<MinimapClickEvent>,
+    mut click_events: EventReader<MinimapPressEvent>,
```
Indy2222, line 165: `click_events` -> `press_events`

crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -63,50 +76,117 @@ impl MinimapClickEvent {
...
 }
 
+fn drag_handler(
+    window_query: Query<&Window, With<PrimaryWindow>>,
+    hud: HudNodes<With<MinimapNode>>,
+    bounds: Res<MapBounds>,
+    dragging: Res<DraggingButtons>,
+    mut drag_events: EventWriter<MinimapDragEvent>,
+) {
+    if dragging.0.is_empty() {
```
Indy2222, line 145: Instead of using `dragging.0` (at multiple places), we can use Bevy's `[derive(Deref)]` https://docs.rs/bevy/latest/bevy/prelude/derive.Deref.html

crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -63,50 +76,117 @@ impl MinimapClickEvent {
...
+
+    let Some(cursor) = window_query.single().cursor_position() else {
+        return;
+    };
+
+    if let Some(relative) = hud.relative_position(cursor) {
+        let proportional = Vec2::new(relative.x, 1. - relative.y);
+        let world = bounds.rel_to_abs(proportional);
+
+        for b in &dragging.0 {
```
Indy2222, line 157: `b` -> `button` would be more readable.

Indy2222 (changes requested):
I've left multiple comments, but nothing major.

crates/controller/src/hud/minimap/interaction.rs
Zakru: At least on Windows, when I click -> alt-tab -> release, I don't seem to be receiving a mouse released event, unless the mouse is over the window.
On the other hand when I simply drag from the window to the outside, I consistently receive a release event, but `cursor_position` is inconsistently sometimes `None` and sometimes `Some` with the correct coordinates outside the window.
Any tricks for this?

Indy2222: What about checking current state (i.e. Bevy's `Input<MouseButton>::.pressed(MouseButton::Left)`) & combining that with flag whether the last press event occurred over the minimap (every time the mouse button is click, we would update en enum of where it was, i.e. `LastPressLocation::Minimap`)?

Zakru: `Input<MouseButton>` ends up having the same shortcomings, as it is internally being updated using the same `MouseButtonInput` events. Seems like we almost need global input if we want to handle this case "perfectly".

crates/controller/src/hud/minimap/interaction.rs
```diff
@@ -63,68 +78,137 @@ impl MinimapClickEvent {
...
+    mut press_events: EventWriter<MinimapPressEvent>,
 ) {
-    let Some(cursor) = window_query.single().cursor_position() else {
-        return;
-    };
+    let cursor = window_query.single().cursor_position();
+
     for event in input_events.iter() {
-        if event.state != ButtonState::Released {
+        if event.state != ButtonState::Pressed {
```
Indy2222, line 123: I would `match event.state` so that it doesn't break when other `ButtonState`s are added in the future.

Indy2222 (changes requested):
One minor note, otherwise LGTM.

crates/controller/src/hud/minimap/interaction.rs
Indy2222: I am ok with merging it as is. Can you please open a new issue so that we fix that later?

Zakru: I was thinking the same.

Zakru:
Opened #715 based on release event issues

Indy2222 (approved):

PR #649, opened 2023-07-28
graph prototype

JackCrumpLeys:
I opened this pr so we could have some discussion over the code as this is a very inf uncial system for the game. As it currently stands we have a NearbyUnits component on every unit that contains producers and receivers (we only have receivers at the moment) this is updates when units move more than a neglagable amount. We then assemble the graph based on changes in NearbyUnits. Next steps include the logic of what to do with the graph and implementing performance operations and maybe a separate tick to just the frame.


Cargo.toml
```diff
@@ -144,6 +146,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 serde_yaml = "0.9"
 sha3 = "0.10.6"
+smallvec = "1.10.0"
```
Indy2222, line 149: We already depend on `tinyvec`, cannot we use that?

Cargo.toml
```diff
@@ -112,6 +112,7 @@ async-compat = "0.2.1"
 async-std = "1.11"
 async-tar = "0.4.2"
 bevy = { version = "0.11", features = ["mp3"] }
+bevy_prototype_debug_lines = { version="0.11", features = ["3d"] }
```
Indy2222, line 115: Can we hide this behind a feature flag (similar to godmode)?

crates/energy/Cargo.toml
```diff
@@ -14,4 +14,13 @@ categories.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+de_spawner.workspace = true
```
Indy2222, line 17: Please separate DE & other dependendics and make the groups sorted. Camera example:

```toml
# DE
de_conf.workspace = true
de_core.workspace = true
de_map.workspace = true
de_terrain.workspace = true
de_uom.workspace = true

# Other
bevy.workspace = true
parry3d.workspace = true
```

crates/energy/Cargo.toml
```diff
@@ -14,4 +14,13 @@ categories.workspace = true
...
+
+
 bevy.workspace = true
+de_core.workspace = true
+smallvec.workspace = true
+de_index.workspace = true
+parry3d.workspace = true
+
+bevy_prototype_debug_lines.workspace = true
```
Indy2222, line 26: Please add a newline here.

crates/energy/src/battery.rs
```diff
@@ -58,6 +63,12 @@ impl Battery {
     }
 }
 
+fn spawn_battery(mut commands: Commands, newly_spawned_units: Query<Entity, Added<Active>>) {
+    for entity in newly_spawned_units.iter() {
+        commands.entity(entity).insert((Battery::default(),));
```
Indy2222, line 68: The battery does not need to be enclosed in a tuple.

crates/energy/src/lib.rs
```diff
@@ -1,14 +1,18 @@
...
 use crate::battery::BatteryPlugin;
 
 pub struct EnergyPluginGroup;
 
 impl PluginGroup for EnergyPluginGroup {
     fn build(self) -> PluginGroupBuilder {
-        PluginGroupBuilder::start::<Self>().add(BatteryPlugin)
+        PluginGroupBuilder::start::<Self>()
+            .add(BatteryPlugin)
+            .add(graph::GraphPlugin)
```
Indy2222, line 16: `graph::GraphPlugin` -> `GraphPlugin` for consistency reasons.

crates/spawner/src/spawner.rs
```diff
@@ -17,10 +16,21 @@ pub(crate) struct SpawnerPlugin;
...
+            spawn
+                .run_if(in_state(GameState::Playing))
+                .in_set(SpawnerSet::Spawn),
+        );
     }
 }
 
+#[derive(SystemSet, Debug, Clone, PartialEq, Eq, Hash)]
+pub enum SpawnerSet {
+    /// This exists so that we can manage newly spawned objects by `.after(SpawnerSet::Spawn)`.
```
Indy2222, line 30: ~I would rather document what is done in the set than what should be done after it (possibly in different creates).~

It is not needed at all, newly spawned entities are not available during the same (`Update`) schedule. Spawn commands are executed only between schedules (e.g. between `Update` and `PostUpdate`).

crates/energy/src/battery.rs
```diff
@@ -1,10 +1,15 @@
...
+use de_spawner::SpawnerSet;
 
 pub(crate) struct BatteryPlugin;
 
 impl Plugin for BatteryPlugin {
     fn build(&self, app: &mut App) {
-        app.add_systems(Update, discharge_battery);
+        app.add_systems(
+            Update,
+            (discharge_battery, spawn_battery.after(SpawnerSet::Spawn)),
```
Indy2222, line 11: Ditto, `spawn_battery()` should be executed during `PostUpdate` as the spawned entity is not available immediately during `Update`. Therefore, `SpawnerSet::Spawn` is not even needed.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,242 @@
...
+    fn into_inner(self) -> SmallVec<[Entity; 40]> {
+        match self {
+            Nearby::Receiver(inner) => inner,
+            Nearby::Producer(inner) => inner,
+        }
+    }
+}
+
+#[derive(Component, Default, Debug, Clone)]
+pub struct NearbyUnits(SmallVec<[Nearby; 2]>, Option<Vec2>);
```
Indy2222, line 87: * Named fields would make the code easier to read.
* Why isn't this `{ procucers: SmallVec<[Entity; 40]>, receivers: SmallVec<[Entity; 40]>, ... }`?

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,242 @@
...
+        .for_each_mut(|(entity, mut nearby_units, transform)| {
+            if let Some(last_pos) = nearby_units.1 {
+                if transform.translation.to_flat().distance_squared(last_pos) < 0.5 {
+                    return;
+                }
+            }
+            nearby_units.1 = Some(transform.translation.to_flat());
+
+            let aabb = &Aabb::new(
+                Point::from(*(transform.translation - Vec3::splat(MAX_DISTANCE)).as_ref()),
```
Indy2222, line 126: `*(...).as_ref()` why is this needed?

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,242 @@
...
+                Point::from(*(transform.translation + Vec3::splat(MAX_DISTANCE)).as_ref()),
+            );
+
+            let producers = spacial_index_producer.query_aabb(aabb, Some(entity));
```
Indy2222, line 130: `query_aabb()` hot that AFAIK this call can be made much faster if needed. There is a lost of space for optimization.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,242 @@
...
+        });
+    println!("update_nearby_recv: {:?}", time.elapsed());
+}
+
+fn update_nearby(
+    mut nearby_units: Mut<NearbyUnits>,
+    producers: Vec<Entity>,
+    receivers: Vec<Entity>,
+) {
+    let mut nearby_producers = SmallVec::new();
```
Indy2222, line 144: Cannot we clear & refill the existing vecs instead of creating new ones?

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,242 @@
...
+    nearby_units.0 = smallvec![
+        Nearby::Producer(nearby_producers),
+        Nearby::Receiver(nearby_receivers)
+    ];
+}
+
+#[allow(clippy::type_complexity)]
+fn update_graph(
+    mut power_grid: ResMut<PowerGrid>,
+    nearby_units: Query<(Entity, &NearbyUnits), Or<(Added<NearbyUnits>, Changed<NearbyUnits>)>>,
```
Indy2222, line 164: [`Chanded<>`](bevy::ecs::prelude::Changed) matches even added entities, therefore the whole `Or<>` can be simplified and the Clippy allow removed.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,242 @@
...
+fn remove_old_nodes(
+    mut power_grid: ResMut<PowerGrid>,
+    mut nearby_units: Query<&mut NearbyUnits>,
+    mut death_events: EventReader<DespawnedComponentsEvent<NearbyUnits>>,
+) {
+    for event in death_events.iter() {
+        power_grid.graph.remove_node(event.entity);
+
+        for outer_nearby in event.data.0.iter().flat_map(|nearby| nearby.clone().into_inner()) {
+            for inner_nearby in nearby_units.get_mut(outer_nearby).unwrap().0.iter_mut(){
```
Indy2222, line 216: This seems a bit convoluted. I would create methods in `Nearby` for removal of entities so that we do not have to call `retain()` in this system. Furthermore, `event.data.0` can have `producers` and `receivers`, these can be iterated so that opposite `producer -> receiver` can be updated in the other entity.

Indy2222 (changes requested):
I have done a first round of review. I will look again with more focus on the details once this gets a bit more mature.

It is important to extensively test the performance. There should be no drops in FPS below 60Hz (even momentary ones) and the CPU usage should stay reasonable. Also, the memory usage must be checked as well.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,242 @@
...
+        commands
+            .entity(entity)
+            .insert((EnergyReceiver, NearbyUnits::default()));
+    }
+}
+
+fn update_nearby_recv(
+    spacial_index_producer: SpatialQuery<Entity, With<EnergyProducer>>,
+    spacial_index_receiver: SpatialQuery<Entity, With<EnergyReceiver>>,
+    mut nearby_units: Query<(Entity, &mut NearbyUnits, &Transform)>,
```
Indy2222, line 112: I would add `, Changed<Transform>` because even the `if transform.translation.to_flat().distance_squared(last_pos) < 0.5 {` is not free.

crates/energy/src/graph.rs
JackCrumpLeys: You mentioned at some point that nearby would be useful for many different reasons (attacking, ai, etc). And by using an enum it means that this can be extended with many different categories. The final resting place for the nearby logic should be somewhere like `de_index`.

crates/energy/src/graph.rs
JackCrumpLeys: It is a vec3 `.as_ref()` turns it into a `&[f64;3]` and the method takes a [f64;3].

crates/energy/src/graph.rs
Indy2222: You are right, my guess is that we will utilize it in other systems as well. However, until we do, I would keep the logic as simple as possible.

crates/energy/src/graph.rs
Indy2222: I think there is `From<Vec3>` implementation for `Point`. https://docs.rs/nalgebra/0.32.3/nalgebra/geometry/type.Point3.html#impl-From%3CVec3%3E-for-Point3%3Cf32%3E-1

JackCrumpLeys: You are correct, I rely on my IDE quite heavily for indexing/finding methods and it has failed in this case.

crates/energy/src/graph.rs
JackCrumpLeys: How about I split this logic into a new PR and put it into de_index?

crates/energy/src/graph.rs
JackCrumpLeys: This can be another PR. I will do some research into optimising this.

Cargo.toml
JackCrumpLeys: The type stored in `TinyVec` must implement `Default`, It is not feasible to do that for Entities, if you insist on using this we could store `Option<Entity>`.

crates/energy/src/graph.rs
JackCrumpLeys: I understand our first point but I'm not sure what you mean for your second point.

crates/energy/src/graph.rs
Indy2222: Yes, the optimizations are separate and should not be done in this PR.

First thing to optimize would be the iteration of candidate entities. The entities are first produced by an iterator and later filtered by the query. We can reverse this with e.g. a closure to make the iterator faster. This optimization alone has the potential to make it much faster on queries matching only a small subset of all entities.

Cargo.toml
Indy2222: Or we can create a wrapper type which implements Default (e.g. `Entity::PLACEHOLDER`).

crates/energy/src/graph.rs
Indy2222: I think that we should keep the logic highly specialized for the energy until we encounter first case where we would utilize a generalized logic. At that point, we can move it to `de_index` and make it somewhat more general. YAGNI

crates/energy/src/graph.rs
Indy2222: The second point is that while iterating over `producers`, you only need to update `receivers` of pointed to entities. And vice versa.

crates/energy/benches/nearby.rs
```diff
@@ -0,0 +1,161 @@
...
+        init_world_with_entities(&mut app.world, *i);
+        app.add_systems(
+            Update,
+            (teleport_entities_randomly, update_nearby_recv, update_index),
+        );
+
+        group.throughput(criterion::Throughput::Elements(*i as u64));
+        group.bench_function(format!("{} entities all teleporting randomly", i), |b| {
+            b.iter(|| {
+                app.update();
```
Indy2222, line 128: This benchmarks (naive) index updating and `update_nearby_recv()` at once. Is it possible to separate this to and benchmark solely `update_nearby_recv()` (i.e. moving and index update would happen outside of the measurement)?

Indy2222 (changes requested):
I did review of the benchmark. How fast is it in current form?

JackCrumpLeys:
> I did review of the benchmark. How fast is it in current form?

If I remember correctly for the moving benchmark: 100: 60 micro seconds, 1000: 200 micro seconds, 10,000: 1.8 ms

This is running on my 16 thread 4gz Ryzen 7 so I recommend you try it on your own gear. If you would like I have access to low spec pcs and could benchmark on a range of hardware.

Indy2222:
> If I remember correctly for the moving benchmark: 100: 60 micro seconds, 1000: 200 micro seconds, 10,000: 1.8 ms

> This is running on my 16 thread 4gz Ryzen 7 so I recommend you try it on your own gear. If you would like I have access to low spec pcs and could benchmark on a range of hardware.

I have AMD Ryzen 7 5800X 8-Core Processor although not overclocked, so it should be pretty similar.

We do not necessarily have to support low specs PCs, but a few year old Desktop CPU with 4-cores should stay well under ~5ms (updates at 60Hz, ~30% CPU usage just for this). I think that we are near that already.

Note that due to the change detection + skipping objects which moved less than 0.5m, there won't be much positive effect from reducing the update frequency. However movement speed will have a big impact so we should set reasonable velocity in the benchmark.

I'll re-run the benchmark once this gets finalized.

crates/energy/benches/nearby.rs
```diff
@@ -0,0 +1,142 @@
...
+        let y: f32 = point.y;
+
+        let collider = LocalCollider::new(
+            ObjectCollider::from(TriMesh::from(Cuboid::new(Vector::new(3., 3., 4.)))),
+            Isometry::new(Vector::new(x, 0., y), Vector::identity()),
+        );
+
+        let entity = world
+            .spawn((
+                Transform::from_xyz(x, y, 0.0),
```
Indy2222, line 72: Should be `x, 0., -y`. But better: `let point_msl = point.to_msl()` (implemented by `ToAltitude`) and use `point_msl` for both the isometry and this transform (`Transform::from_translation(point_msl)`).

crates/energy/benches/nearby.rs
```diff
@@ -0,0 +1,142 @@
...
+            -1.
+        };
+
+        let t = time.elapsed_seconds();
+        let radius = DISTANCE_FROM_MAP_EDGE / 2.;
+
+        let x = radius * (t * 0.5 * direction).sin();
+        let y = radius * (t * 0.5 * direction).cos();
+
+        transform.translation.x = x;
```
Indy2222, line 101: This should be `center.x + x` (each entity needs to have a center) so that all entities doesn't circle around the same center added by `init_world_with_entities_moving`.

crates/energy/benches/nearby.rs
```diff
@@ -0,0 +1,152 @@
...
+    for (mut transform, unit_number, centre) in query.iter_mut() {
+        // Change direction (counter)clockwise based on entity_mum % 2 == 0
+        let direction = if unit_number.0 as f32 % 2. == 0. {
+            1.
+        } else {
+            -1.
+        };
+
+        let t = clock.0;
+        let radius = DISTANCE_FROM_MAP_EDGE / 2.;
```
Indy2222, line 103: We do not have to divide this by two since we keep the center and more at most `radius` from it. Also, I would rename it to `MOVEMENT_RADIUS` which semantically fits more the purpose of the variable (we keep the distance from map edge this large due to the radius of the circular movement and not vice versa).

Note that for a small radius and a large map, accompanying for the radius via the margin has a minimal practical effect. But its more "correct".

crates/energy/benches/nearby.rs
```diff
@@ -0,0 +1,152 @@
...
+        );
+        index.update(entity, position);
+    }
+}
+
+fn init_world_with_entities_moving(world: &mut World, num_entities: u32) {
+    let mut index = EntityIndex::new();
+    let points = load_points(
+        num_entities.try_into().unwrap(),
+        MAP_SIZE - DISTANCE_FROM_MAP_EDGE,
```
Indy2222, line 58: We should multiply by two here, since there are two margins (one on each side). This is mathematically corrected by the division by two elsewhere (see my comment there) but semantically, the correction belongs here.

Also, I would store the result to a variable and assert that its greater than 0.

crates/energy/benches/nearby.rs
```diff
@@ -0,0 +1,152 @@
...
+    for i in [100, 1_000, 10_000, 100_000].iter() {
+        let mut app = App::default();
+        init_world_with_entities_moving(&mut app.world, *i);
+        app.add_systems(Update, update_nearby_recv);
+        app.add_plugins(TimePlugin);
+
+        let update_units_schedule = Schedule::default();
+        app.add_schedule(UpdateUnits, update_units_schedule);
+
+        app.add_systems(UpdateUnits, (update_index, move_entities_in_circle));
```
Indy2222, line 127: I would run this two systems in chain so that the transform update **always** happens before the index update.

crates/energy/benches/nearby.rs
```diff
@@ -0,0 +1,152 @@
...
+                let time = std::time::Instant::now();
+                let mut duration_updating_other_stuff = std::time::Duration::default();
+
+                for _ in 0..iters {
+                    let update_other_stuff = std::time::Instant::now();
+                    app.world.resource_mut::<Clock>().inc();
+                    app.world.run_schedule(UpdateUnits);
+                    duration_updating_other_stuff += update_other_stuff.elapsed();
+
+                    app.update();
```
Indy2222, line 141: Doesn't this run both schedules?

crates/energy/benches/nearby.rs
```diff
@@ -0,0 +1,152 @@
...
+
+        group.throughput(criterion::Throughput::Elements(*i as u64));
+        group.bench_function(format!("{} units all moving in circles", i), |b| {
+            b.iter_custom(|iters| {
+                let time = std::time::Instant::now();
+                let mut duration_updating_other_stuff = std::time::Duration::default();
+
+                for _ in 0..iters {
+                    let update_other_stuff = std::time::Instant::now();
+                    app.world.resource_mut::<Clock>().inc();
```
Indy2222, line 137: I would move this line above the line above so that it is not included in the measurement.

Indy2222 (changes requested):
It is looking good :heart:. I only left mostly very technical comments, nothing major.

Indy2222 (changes requested):

crates/energy/benches/nearby.rs
JackCrumpLeys: No just the Main one, see docs:
> Advances the execution of the Schedule by one cycle.
This method also updates sub apps. See insert_sub_app for more details.
***The schedule run by this method is determined by the main_schedule_label field. By default this is Main.***

crates/energy/benches/nearby.rs
JackCrumpLeys: Why? this way any cost caused by it is eliminated by the benchmark, isolating the nearby system.

Indy2222: Exactly, you want to measure just the nearby system not the clock incrementing. OTOH, the impact would likely be negligible.

crates/energy/benches/nearby.rs
```diff
@@ -93,12 +97,12 @@ fn move_entities_in_circle(
         let direction = if unit_number.0 % 2 == 0 { 1. } else { -1. };
 
         let t = clock.0;
-        let radius = DISTANCE_FROM_MAP_EDGE / 2.;
+        let radius = MOVEMENT_RADIUS;
```
Indy2222, line 100: This variable is a bit redundant, you can use the constant directly.

crates/energy/benches/nearby.rs
```diff
@@ -0,0 +1,161 @@
...
+    for (mut transform, unit_number, centre) in query.iter_mut() {
+        // Change direction (counter)clockwise based on entity_mum % 2 == 0
+        let direction = if unit_number.0 % 2 == 0 { 1. } else { -1. };
+
+        let t = clock.0;
+        let radius = MOVEMENT_RADIUS;
+        let omega = SPEED / radius;
+        let omega_shift = 0.1 * unit_number.0 as f32;
+
+        let x = radius * (t * omega_shift + omega * direction).sin();
```
Indy2222, line 104: Should be `omega_shift + t * omega * direction`

crates/energy/benches/nearby.rs
JackCrumpLeys: I'm confused, if it is moved above the timer that times the update time for unrelated stuff then it would incur cost to the final benchmark. At its current placement it does not.

crates/energy/benches/nearby.rs
```diff
@@ -53,8 +53,8 @@ fn update_index(
 
 fn init_world_with_entities_moving(world: &mut World, num_entities: &NumPoints) {
     let mut index = EntityIndex::new();
-    let max_point_value = MAP_SIZE - MOVEMENT_RADIUS * 2.;
-
+    let max_point_value =
+        UNIT_SPACING * (<NumPoints as Into<usize>>::into(*num_entities) as f32).sqrt();
```
Indy2222, line 57: `num_entities.into::<usize>()` should work as well.

crates/energy/benches/nearby.rs
Indy2222: You are right, I got confused. We measure time of other stuff not the bench-marked stuff.

crates/energy/benches/nearby.rs
```diff
@@ -149,9 +149,17 @@ fn nearby_benchmark(c: &mut Criterion) {
...
-                })
+                });
             },
         );
+        let mut amount_of_connections_formed = 0;
+        for connections in app.world.query::<&mut NearbyUnits>().iter(&app.world) {
+            amount_of_connections_formed += connections.units().len()
+        }
+        println!(
+            "We ended up with {} connections between {} units",
```
Indy2222, line 160: Note that every connection is counted twice.

crates/energy/benches/nearby.rs
JackCrumpLeys: That is not how into works. Into is a function with no type parameter of a struct implementing `Into<T>` and returning `T`. I did give it a go anyway and it did not compile.

crates/energy/benches/nearby.rs
JackCrumpLeys: Connections are two ways and that is subject to change but I think that is fine. OTOH we could assemble the graph as well and benchmark that and count edges if you would like.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+pub struct EnergyReceiver;
+
+/// The energy producer component is used to mark an entity as an energy producer.
+#[derive(Component, Debug, Clone, Copy)]
+pub struct EnergyProducer;
+
+/// The nearby component is used to store the nearby entities of an entity.
+#[derive(Debug, Clone, Default)]
+pub struct Nearby {
+    receivers: TinyVec<[NearbyEntity; 32]>,
```
Indy2222, line 103: My guess is that array of size ~8 would be a performance sweet spot.

Too little and most of the vecs go to the heap, too many and the component becomes large and thus cache inefficient (they are stored in packed arrays by the Bevy ECS).

Since we have a benchmark, it should be easy to test (on large number of entities) where is the sweet spot. Although, that would be for the location distribution (uniform random) in the benchmark. In reality, it would be more clustered.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+    producers: TinyVec<[NearbyEntity; 32]>,
+}
+
+impl Nearby {
+    fn both(&self) -> TinyVec<[NearbyEntity; 64]> {
+        self.receivers
```
Indy2222, line 109: I would rather return an iterator for performance reason (no need to construct a new `TinyVec` every time this is called). Alternatively, it can take a `&mut Vec` as an argument.

Note that I suggest to not separate producers & receivers (i.e. to have only single vec) in another comment. If we go that way, the method would not be needed at all.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+const MAX_DISTANCE: f32 = 10.0;
+
+pub(crate) struct GraphPlugin;
+
+impl Plugin for GraphPlugin {
+    fn build(&self, app: &mut App) {
+        app.add_plugins(DespawnEventsPlugin::<&NearbyUnits, NearbyUnits>::default())
+            .add_systems(OnEnter(GameState::Playing), setup)
+            .add_systems(OnExit(GameState::Playing), clean_up)
+            .add_systems(PostUpdate, spawn_graph_components)
```
Indy2222, line 28: Should be enough to run this just in `AppState::InGame`.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+
+// The max distance (in meters) between two entities for them to be consider neighbors in the graph
+const MAX_DISTANCE: f32 = 10.0;
+
+pub(crate) struct GraphPlugin;
+
+impl Plugin for GraphPlugin {
+    fn build(&self, app: &mut App) {
+        app.add_plugins(DespawnEventsPlugin::<&NearbyUnits, NearbyUnits>::default())
+            .add_systems(OnEnter(GameState::Playing), setup)
```
Indy2222, line 26: I think we should do the setup and cleanup on `GameState::InGame`. The resources / components might be useful even during e.g. game loading.

Note that the states are hierarchical. `AppState` is top-level state, `GameState` divides `AppState::InGame` to multiple "sub-states".

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+            .add_systems(
+                FixedUpdate,
+                (
+                    remove_old_nodes.before(GraphSystemSet::UpdateNearby),
+                    update_nearby_recv.in_set(GraphSystemSet::UpdateNearby),
+                    update_graph
+                        .in_set(GraphSystemSet::UpdateGraph)
+                        .after(GraphSystemSet::UpdateNearby),
+                )
+                    .run_if(in_state(GameState::Playing)),
```
Indy2222, line 38: It should run whole `AppState::InGame` because game loading phase when map entities are spawned make take longer than a single frame (especially during multiplayer).

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+}
+
+fn clean_up(mut commands: Commands) {
+    commands.remove_resource::<PowerGrid>();
+}
+
+/// This system spawns Energy Producers and Energy Receivers and nearby units.
+fn spawn_graph_components(
+    mut commands: Commands,
+    newly_spawned_units: Query<Entity, Added<Active>>,
```
Indy2222, line 150: IMHO semantically proper way would be to trigger this on `Battery` rather than `Active`. Id est, anything with a battery can receive energy to charge it.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+pub fn update_nearby_recv(
+    spacial_index_producer: SpatialQuery<Entity, With<EnergyProducer>>,
+    spacial_index_receiver: SpatialQuery<Entity, With<EnergyReceiver>>,
+    mut nearby_units: Query<(Entity, &mut NearbyUnits, &Transform), Changed<Transform>>,
+) {
+    nearby_units
+        .par_iter_mut()
+        .for_each_mut(|(entity, mut nearby_units, transform)| {
+            if let Some(last_pos) = nearby_units.last_pos {
+                if transform.translation.to_flat().distance_squared(last_pos) < 0.5 {
```
Indy2222, line 168: I would store `transform.translation.to_flat()` to a variable since it is used twice.

Indy2222, line 168: I would move `0.5` to a constant with a documentation comment: `Minimum distance squared traveled by an object to update its nearby units.`

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+                }
+            }
+            nearby_units.last_pos = Some(transform.translation.to_flat());
+
+            let aabb = &Aabb::new(
+                Point::from(transform.translation - Vec3::splat(MAX_DISTANCE)),
+                Point::from(transform.translation + Vec3::splat(MAX_DISTANCE)),
+            );
+
+            let producers = spacial_index_producer.query_aabb(aabb, Some(entity));
```
Indy2222, line 179: The code for producers and receivers is almost identical. I would try to move it to a function and call it with arguments `index, entities`.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+
+            let receivers = spacial_index_receiver.query_aabb(aabb, Some(entity));
+
+            nearby_units.units.clear();
+
+            nearby_units
```
Indy2222, line 185: I think it might be slightly more efficient to do the spatial query and update the entities before doing the second spatial query (i.e. not having the code interleaved).

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+fn remove_old_nodes(
+    mut power_grid: ResMut<PowerGrid>,
+    mut nearby_units: Query<&mut NearbyUnits>,
+    mut death_events: EventReader<DespawnedComponentsEvent<NearbyUnits>>,
+) {
+    for event in death_events.iter() {
+        power_grid.graph.remove_node(event.entity);
+
+        // Remove the entity from the nearby units of all nearby units
+        for mut nearby_units in nearby_units.iter_mut() {
```
Indy2222, line 240: It is not necessary to update all of the. It would be enough (pseudocode):

```rust
let despawned_nearby = nearby_units.get(event.entity);
for producer in despawned_nearby.producers {
    nearby_units.get_mut(producer).receivers.remove(event.entity);
}
for receiver in despawned_nearby.receivers {
    nearby_units.get_mut(receiver).producers.remove(event.entity);
}
```

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+    }
+}
+
+#[cfg(feature = "energy_graph_debug_lines")]
+fn debug_lines(
+    power_grid: Res<PowerGrid>,
+    query: Query<&Transform>,
+    mut debug_lines: ResMut<DebugLines>,
+) {
+    for (node, _) in power_grid.graph.node_references() {
```
Indy2222, line 252: Doesn't this add every edge twice (since both end nodes are iterated over)? I would instead use [all_edges](https://docs.rs/petgraph/latest/petgraph/graphmap/struct.GraphMap.html#method.all_edges)

crates/energy/src/graph.rs
Indy2222, line 185: Originally we discussed the need to have separate producers and receivers. Thinking about it more, we might use only one vec and allow flow of energy in any direction.

Reasoning: 

* It would simplify the implementation.
* It would better mimic real world (e.g. you might charge your EV from the wall and sell energy from the battery back to the grid when you don't need to use your EV).

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+                .producers
+                .extend(producers.map(|entity| entity.into()));
+            nearby_units
+                .units
+                .receivers
+                .extend(receivers.map(|entity| entity.into()));
+        });
+}
+
+fn update_graph(
```
Indy2222, line 196: As I suggested in another comment, I would leave actual graph implementation to another PR.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+            &mut power_grid
+                .graph
+                .edges(entity)
+                .map(|edge| (edge.source(), edge.target())),
+        );
+
+        let mut edges_to_add = vec![];
+
+        for nearby_entity in nearby_units.units.both() {
+            edges_to_add.push((entity, nearby_entity.into()));
```
Indy2222, line 213: No need to create `edges_to_add`, it is used only once so you can iterate `nearby_units.units.both()` directly.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+                .extend(receivers.map(|entity| entity.into()));
+        });
+}
+
+fn update_graph(
+    mut power_grid: ResMut<PowerGrid>,
+    nearby_units: Query<(Entity, &NearbyUnits), Changed<NearbyUnits>>,
+) {
+    for (entity, nearby_units) in nearby_units.iter() {
+        let mut edges_to_remove = HashSet::new();
```
Indy2222, line 201: Using HashSet is less efficient than using an array, tinyvec or vec for small number of entities (possible up to several dozens).

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,259 @@
...
+
+        edges_to_remove.extend(
+            &mut power_grid
+                .graph
+                .edges(entity)
+                .map(|edge| (edge.source(), edge.target())),
```
Indy2222, line 207: I would use `filter_map` and check whether the edge is in `nearby_units.units.both()` so that we do not have to remove from the set later.

Indy2222 (changes requested):
:tada: we are getting close to a functional energy grid. I am super exited about that!

Cargo.toml
```diff
@@ -135,6 +137,7 @@ ntest = "0.9.0"
 parry2d = "0.13.1"
 parry3d = "0.13.1"
 paste = "1.0.12"
+petgraph = "0.6.3"
```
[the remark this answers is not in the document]
JackCrumpLeys: I think Im going to close this pr then split the branch into several prs.

crates/energy/Cargo.toml
```diff
@@ -13,5 +13,32 @@ categories.workspace = true
...
+de_index.workspace = true
+de_spawner.workspace = true
+
+# Other
 bevy.workspace = true
+parry3d.workspace = true
+tinyvec.workspace = true
+
+# This cannot be a workspace dependency because it is optional
+bevy_prototype_debug_lines = { workspace = true, optional = true }
```
Indy2222, line 31: I would keep it sorted with the "Other" dependencies.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,239 @@
...
+            .add_systems(OnEnter(AppState::InGame), setup)
+            .add_systems(OnExit(AppState::InGame), clean_up)
+            .add_systems(
+                PostUpdate,
+                spawn_graph_components.run_if(in_state(AppState::InGame)),
+            )
+            .add_systems(
+                Update,
+                remove_old_nodes
+                    .before(GraphSystemSet::UpdateNearby)
```
Indy2222, line 37: .before() is unnecessary because it happens during a different schedule.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,239 @@
...
+/// The nearby units component is used to store the nearby entities of an entity.
+#[derive(Component, Default, Debug, Clone)]
+pub struct NearbyUnits {
+    units: TinyVec<[NearbyEntity; 16]>,
+    last_pos: Option<Vec2>,
+}
+
+impl NearbyUnits {
+    fn remove_matching(&mut self, entity: NearbyEntity) {
+        self.units.retain(|e| *e != entity);
```
Indy2222, line 113: Finding index of the entity and using swap-remove might be faster. Retain has to shift all following entities and thus is somewhat slower.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,239 @@
...
+        println!("removed {:?} from {:?}", entity, self.units);
```
Indy2222, line 114: Reminder: do not forget to remove this debug print.

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,239 @@
...
+
+fn setup(mut commands: Commands) {
+    commands.insert_resource(PowerGrid::default());
+}
+
+fn clean_up(mut commands: Commands) {
+    commands.remove_resource::<PowerGrid>();
+}
+
+/// This system spawns Energy Producers and Energy Receivers and nearby units.
```
Indy2222, line 134: Please update the doc comment (receivers & producers were unified).

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,239 @@
...
+                }
+            }
+            nearby_units.last_pos = Some(current_pos);
+
+            let aabb = &Aabb::new(
+                Point::from(transform.translation - Vec3::splat(MAX_DISTANCE)),
+                Point::from(transform.translation + Vec3::splat(MAX_DISTANCE)),
+            );
+
+            let members = spacial_index_member.query_aabb(aabb, Some(entity));
```
Indy2222, line 166: I would swap this and the following lines & remove the empty lines:

```rust
nearby_units.units.clear();
let members = spacial_index_member.query_aabb(aabb, Some(entity));
nearby_units.units.extend(members.map(NearbyEntity));
```

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,239 @@
...
+
+            nearby_units.units.clear();
+
+            nearby_units.units.extend(members.map(NearbyEntity));
```
Indy2222, line 170: This system updates only entities which moved by some minimal distance -> there must be reciprocal update of non-moving entities which got close / far from neighbors do the the movement of neighbors.

Possible solution:
* get difference between past and future nearby units,
* add self to all newly added units,
* remove self from all newly removed units,
* update units on self.

Indy2222 (changes requested):
I reviewed all but `update_graph()` system. I think there is a bug in how we update nearby units.

Indy2222 (changes requested):

crates/energy/src/graph.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    pub fn is_empty(&self) -> bool {
+        self.units.is_empty()
+    }
+}
+
+/// This system inserts newly spawned units into the energy grid.
+fn spawn_graph_components(
+    mut commands: Commands,
+    newly_spawned_units: Query<Entity, Added<Battery>>,
```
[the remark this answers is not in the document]
JackCrumpLeys: That was the original code see: https://github.com/DigitalExtinction/Game/pull/649#discussion_r1314310563. I think it would be a better idea just to fix the system order as its more logical that each battery has a connection to the graph.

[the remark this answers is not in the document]
Indy2222: LOL, I am disputing my own idea. Now it leads to the delay so I would try to solve it one way or another (possibly with an explanatory comment why it is needed).

PR #738, opened 2023-09-19
test utils

JackCrumpLeys:
unify test data for better deterministic benchmarks and testing.


crates/index/benches/ray.rs
```diff
@@ -87,8 +65,8 @@ fn setup_world(world: &mut World, num_entities: u32, max_distance: f32) {
     }
 
     let mut rays = Rays::new();
-    let ray_origins = load_points(1000);
-    let ray_dirs = load_points(1000);
+    let ray_origins = load_points(&1000.try_into().unwrap(), MAP_SIZE);
```
JackCrumpLeys, line 68: should we just use the enum variant here?

Indy2222: I would.

crates/test_utils/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+            100_000 => Ok(Self::OneHundredThousand),
+            _ => Err("Invalid number of points"),
+        }
+    }
+}
+
+impl From<NumPoints> for usize {
+    fn from(value: NumPoints) -> Self {
+        match value {
+            NumPoints::OneHundred => 100,
```
Indy2222, line 33: I would replace this with `Self::from(&value)` so that it stays DRY.

Indy2222 (changes requested):
Two minor notes, otherwise LGTM.

Indy2222 (approved):

PR #766, opened 2023-10-12
basic keybind logic

### LINKED ISSUE #648: Implement better input system
Currently user inputs (e.g. keyboard presses, mouse inputs) and resulting actions / commands are coupled. Many actions / commands are executed by systems whose run conditions are the specific inputs.

Change this by creating a generic input system. User inputs (mouse drags, moves, clicks, keyboard presses, etc.) result in events. These events are mapped to user commands.

This is a stepping stone for hotkey configuration.

Research how this is done in other games and other systems.

JackCrumpLeys:
related to #648 

- [x] action registering and dispatching based on input key-binds
- [x] Migrate old code to new system
- [ ] ~using a config file to deserialise and serialise user defined key-binds~
- [ ] ~Bubble up config errors (we just panic at the moment)~


Indy2222:
I hoped that we will be able to utilize [leafwing-input-manager](https://github.com/Leafwing-Studios/leafwing-input-manager) instead of doing all the work from scratch.

Notes about how we can structure the usage of leafwing-input-manager:

* leafwing-input-manager (InputManagerPlugin) puts all relevant systems in PreUpdate schedule. We need to move that to `Input` and make sure that a) there is no one frame delay between raw inputs from bevy and when leafwing-input-manager processes them, b) handling of the actions is done after the relevant systems from leafwing-input-manager run.
* All input code lives in controller (which is intended to contain all user control code).
  * This crate contains a bit more than that and we should fix that in the future. However, it is an independent task.
* There is an enum (a single one for now?) with all keyboard/mouse actions (implementing `leafwing_input_manager::Actionlike`).
  * In later stages, this would be (de)serializable so it can be used in config.
  * InputMap from leafwing_input_manager already is (de)serializable.
* During startup, InputMap and ActionState are created.
  * These should be global (we do not need per-entity control).
  * In later stages, the input map would be loaded from configuration.
* InputManagerPlugin<OurActionEnum> is registered or we add the relevant systems manually so that they can be executed during Input schedule.
  * We might ask people from leafwing-input-manager whether it is possible to add a bit more scheduling flexibility to leafwing-input-manager.
* `ClashStrategy::UseActionOrder` is used - to reconsider.

crates/input/src/io.rs
```diff
@@ -0,0 +1,39 @@
+use std::io::Read;
+
+use de_core::fs::conf_dir;
+use leafwing_input_manager::prelude::InputMap;
+use ron::ser::PrettyConfig;
+
+use crate::BindableActionlike;
+
+pub(crate) fn get_keybindings<A: BindableActionlike>(
```
Indy2222, line 9: I would not introduce any (de)serialization in this PR. Lets first migrate to leafwing_input_manager to keep the PR lean.

crates/controller/src/actions.rs
```diff
@@ -0,0 +1,115 @@
...
+    Up, (KeyCode::W, KeyCode::Up),
+    Down, (KeyCode::S, KeyCode::Down),
+    Left, (KeyCode::A, KeyCode::Left),
+    Right, (KeyCode::D, KeyCode::Right),
+    Pivot, (UserInput::Chord(PetitSet::from_iter(vec![KeyCode::ControlLeft.into(), MouseButton::Middle.into()])));
+    // --- mouse actions (these will trigger the drag logic) ---
+    PrimaryClick, (MouseButton::Left),
+    SecondaryClick, (
+        MouseButton::Right);
+    //  --- building actions ---
```
Indy2222, line 101: I would merge this to keyboard actions. It can be just `BuildBase, (KeyCode::B)` and so on.

crates/controller/src/actions.rs
```diff
@@ -0,0 +1,115 @@
...
+impl bevy::app::Plugin for ActionPlugin {
+    fn build(&self, app: &mut App) {
+        app.add_action_set::<Action>("actions")
+            // Mouse is separate because otherwise it will clash with AddToSelection and ReplaceSelection
+            .add_action_set::<MouseAction>("mouse_actions");
+    }
+}
+
+/// make actoinlike enum that has normal actions and factory actions. an action is A vareint followed by a KeyConfig<A>
+macro_rules! make_actions {
```
Indy2222, line 20: I would get rid of this macro and implement all the stuff directly. It is used only once so IMHO it is an unnecessary indirection.

crates/controller/src/actions.rs
```diff
@@ -0,0 +1,115 @@
...
+        UserInput::Chord(PetitSet::from_iter(vec![KeyCode::ControlRight.into(), MouseButton::Left.into()]))),
+    ReplaceSelection, (MouseButton::Left),
+    // camera controls
+    Up, (KeyCode::W, KeyCode::Up),
+    Down, (KeyCode::S, KeyCode::Down),
+    Left, (KeyCode::A, KeyCode::Left),
+    Right, (KeyCode::D, KeyCode::Right),
+    Pivot, (UserInput::Chord(PetitSet::from_iter(vec![KeyCode::ControlLeft.into(), MouseButton::Middle.into()])));
+    // --- mouse actions (these will trigger the drag logic) ---
+    PrimaryClick, (MouseButton::Left),
```
Indy2222, line 98: I think that mouse actions and keyboard actions should be joined. If am not mistaken, setting this to e.g. `MouseButton::Right` would break e.g. `ReplaceSelection` (it implicitly assumes that the two are the same).

I would get rid of `PrimaryClick` completely. Ideally there is a single map of actions to all inputs which trigger the action.

There might be multiple actions triggered by the same input (and handled based on context, for example mouse position). Hopefully, that is possible with the input manager.

Also, I would delete things like `PrimaryClick`. All the actions should be concrete things (like BuildBase, PivotCamera, ...). Also, is it possible to treat doubleclicks properly (something like `MouseButton::DoubleLeft`) with the input manager so we do not have to hack around?

Note that mouse input handling might need to be changed to accommodate for these suggestions.

Indy2222 (changes requested):

PR #767, opened 2023-10-12
de_messages: Remove an unnecessary cfg()

Indy2222:


Indy2222:
This change is not desired because to lead to an unused import.

Issue #354, opened 2023-01-27
Error on Mac using mouse


Polostor:
After upgrading to the latest `main` I have found a bug where I can't use my mouse.
It is stuck in first loaded place and does not move even when it should..

This issue appears on both M1 Mac as well as on Intel Mac.

I have checked and I can confirm, that this does not happen before the PR https://github.com/DigitalExtinction/Game/pull/352

Polostor:
I have done another research and the issue is purely in commit https://github.com/DigitalExtinction/Game/pull/352/commits/5e45926a3fa497e5fc0c9493155bf6d91c585e01

Indy2222:
Probably result of https://github.com/rust-windowing/winit/issues/1093

Indy2222:
#355 implements a stop-gab solution of disabling mouse grabbing on MacOS.

PR #777, opened 2023-12-12
make window mode changeable

### LINKED ISSUE #570: Support non-fullscreen mode
Add support for non-fullscreen:

- [ ] make the initial state configurable
- [ ] switch between a window and fullscreen with F11

JackCrumpLeys:
closes #570 

I think waiting until #766 is closed so that we can do the toggle full-screen keybinding properly is a good idea.


Cargo.toml
```diff
@@ -167,28 +167,29 @@ urlencoding = "2.1.2"
...
-    "ktx2",
-    "zstd",
-    "x11",
-    "filesystem_watcher",
-    "bevy_gizmos",
-    "android_shared_stdcxx",
-    "tonemapping_luts",
-    "default_font",
-    "webgl2",
+  "animation",
```
Indy2222, line 170: Can you revert the reformatting change (keeping just addition of serialize)?

crates/conf/src/conf.rs
```diff
@@ -220,9 +233,17 @@ impl AudioConf {
     }
 }
 
+impl Window {
+    /// Window mode.
```
Indy2222, line 237: This doc comment is a bit redundant IMO.

crates/conf/src/conf.rs
```diff
@@ -220,9 +233,17 @@ impl AudioConf {
...
+    }
+}
+
 // Bundle configuration neatly into a single struct
 bundle_config!(
     camera: CameraConf: Camera, // Conf file -> Camera -> CameraConf
     multiplayer: MultiplayerConf: MultiplayerConf,  // Conf file -> MultiplayerConf
-    audio: AudioConf: AudioConf
+    audio: AudioConf: AudioConf,
+    window: Window: Window
```
Indy2222, line 248: I would place a comma at the end of line (if the macro allows) so that in the future, the diff has only one line (unlike now).

crates/controller/src/window.rs
```diff
@@ -0,0 +1,44 @@
...
+    window::{CursorGrabMode, WindowMode},
+};
+use de_conf::Configuration;
+use de_core::state::AppState;
+
+pub struct WindowManagementPlugin;
+
+impl Plugin for WindowManagementPlugin {
+    fn build(&self, app: &mut App) {
+        app.add_systems(
```
Indy2222, line 12: I think that `add_systems()` returns app, so it is possible to chain these like this:

```rust
app.add_systems(
    OnExit(AppState::AppLoading),
    update_window_mode_based_on_config,
).add_systems(Update, toggle_fullscreen);
```

crates/controller/src/window.rs
```diff
@@ -0,0 +1,44 @@
...
+
+fn toggle_fullscreen(mut windows: Query<&mut Window>, keyboard_input: Res<Input<KeyCode>>) {
+    if keyboard_input.just_pressed(KeyCode::F11) {
+        let window = &mut windows.single_mut();
+        match window.mode {
+            WindowMode::BorderlessFullscreen => {
+                window.mode = WindowMode::Windowed;
+                window.cursor.grab_mode = CursorGrabMode::None;
+            }
+            WindowMode::Windowed => {
```
Indy2222, line 34: This branch is redundant to the catch all branch below.

crates/controller/src/window.rs
```diff
@@ -0,0 +1,44 @@
...
+                window.mode = WindowMode::BorderlessFullscreen;
+                window.cursor.grab_mode = CursorGrabMode::Confined;
```
Indy2222, line 36: This doesn't work on MacOS, see `cursor_grab_system()` and #354. We should move the code here, with `None` on Mac and `Confined` otherwise (with conditional compilation).

Indy2222 (changes requested):
Cool! I have several remarks, but nothing hard to fix.

PR #775, opened 2023-11-09
Upgrade to Bevy v0.12

Indy2222:
- [x] upgrade dependencies
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#bevy-asset-v2
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#simplify-parallel-iteration-methods
- [ ] https://bevyengine.org/learn/migration-guides/0.11-0.12/#fix-safety-invariants-for-worldquery-fetch-and-simplify-cloning
- [ ] https://bevyengine.org/learn/migration-guides/0.11-0.12/#replaced-entitymap-with-hashmap
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#replaced-entitycommand-implementation-for-fnonce
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#move-schedule-name-into-schedule
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#refactor-eventreader-iter-to-read
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#replace-intosystemsetconfig-with-intosystemsetconfigs
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#rename-removedcomponents-iter-iter-with-id-to-read-read-with-id
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#remove-states-variants-and-remove-enum-only-restriction-its-derive
- [ ] https://bevyengine.org/learn/migration-guides/0.11-0.12/#replace-all-labels-with-interned-labels
- [ ] https://bevyengine.org/learn/migration-guides/0.11-0.12/#add-configure-schedules-to-app-and-schedules-to-apply-schedulebuildsettings-to-all-schedules
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#split-computedvisibility-into-two-components-to-allow-for-accurate-change-detection-and-speed-up-visibility-propagation
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#pbr-shader-cleanup
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#allow-extensions-to-standardmaterial
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#update-shader-imports
- [x] https://bevyengine.org/learn/migration-guides/0.11-0.12/#bind-group-entries
[...]

PR #784, opened 2024-02-21
Improve battery discharge unittest

Indy2222:
Now it runs 1s faster and is no longer timing sensitive.


Indy2222:
I tried to backport the fix to the test from #775. However, time in Bevy v0.11 works differently and its not worth the patch noise. Closing.

PR #795, opened 2024-06-16
Add basic enemy colors to minimap

### LINKED ISSUE #584: Draw entities of each player with different color on the minimap
Currently, playable player entities are drawn with one color and all remaining entities in a different color. This makes it impossible to distinguish who is who on the minimap.

Change this by deriving the color from the player (number). This will make the coloring consistent across users during a multiplayer game.

redbugz:
Fixes #584.

All players are assigned a color based on their player number. It also changes the color of the units slightly to distinguish them from buildings.

In the future I think it would be nice that each player could choose their color from a list of curated colors that are visually distinctive, and then they would always be consistent across single player or multiplayer views. Or they are assigned a color, but then the UI would need to change in other ways so they would know which color they were assigned. This could be as simple as a color bar next to the their name in the game list, or some other color indicator in the HUD.

I'm just learning Rust and Bevy, so hopefully I did things properly. Let me know if you would like any changes.

### DIFF
--- a/crates/controller/src/hud/minimap/fill.rs
+++ b/crates/controller/src/hud/minimap/fill.rs
@@ -1,3 +1,6 @@
+use super::draw::DrawingParam;
+use crate::ray::ScreenRay;
+use ahash::AHashMap;
 use bevy::{ecs::system::SystemParam, prelude::*};
 use de_core::{
     gamestate::GameState, gconfig::GameConfig, objects::ObjectTypeComponent,
@@ -6,21 +9,55 @@ use de_core::{
 use de_map::size::MapBounds;
 use de_objects::SolidObjects;
 use de_terrain::TerrainCollider;
-use de_types::projection::ToFlat;
+use de_types::{objects::*, projection::ToFlat};
 use parry2d::{
     bounding_volume::Aabb,
     math::Point,
     query::{Ray, RayCast},
 };
 
-use super::draw::DrawingParam;
-use crate::ray::ScreenRay;
-
 const TERRAIN_COLOR: Color = Color::rgb(0.61, 0.46, 0.32);
 const PLAYER_COLOR: Color = Color::rgb(0.1, 0.1, 0.9);
-const ENEMY_COLOR: Color = Color::rgb(0.9, 0.1, 0.1);
+const ENEMY_COLORS: [Color; 4] = [
+    Color::rgb(0.1, 0.9, 0.1),
+    Color::rgb(0.9, 0.1, 0.1),
+    Color::rgb(0.9, 0.9, 0.1),
+    Color::rgb(0.1, 0.9, 0.9),
+];
 const MIN_ENTITY_SIZE: Vec2 = Vec2::splat(0.02);
 const CAMERA_COLOR: Color = Color::rgb(0.9, 0.9, 0.9);
+
+#[derive(Resource, Debug)]
+struct PlayerColors {
+    player: (Color, AHashMap<ObjectType, Color>),
+    enemies: [(Color, AHashMap<ObjectType, Color>); 4],
+}
+
+impl Default for PlayerColors {
+    fn default() -> Self {
+        Self {
+            player: derive_colors(PLAYER_COLOR),
+            enemies: [
+                derive_colors(ENEMY_COLORS[0]),
+                derive_colors(ENEMY_COLORS[1]),
+                derive_colors(ENEMY_COLORS[2]),
+                derive_colors(ENEMY_COLORS[3]),
+            ],
+        }
+    }
+}
+
+fn derive_colors(base_color: Color) -> (Color, AHashMap<ObjectType, Color>) {
+    let mut unit_color = base_color.clone();
+    unit_color.set_s(base_color.s() - 0.3);
+    let mut object_colors = AHashMap::new();
+    object_colors.insert(
+        ObjectType::Active(ActiveObjectType::Unit(UnitType::Attacker)),
+        unit_color,
+    );
+
+    (base_color, object_colors)
+}
 
 pub(super) struct FillPlugin;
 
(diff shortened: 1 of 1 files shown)


crates/controller/src/hud/minimap/fill.rs
```diff
@@ -6,22 +9,56 @@ use de_core::{
...
+    Color::rgb(0.1, 0.9, 0.1),
+    Color::rgb(0.9, 0.1, 0.1),
+    Color::rgb(0.9, 0.9, 0.1),
+    Color::rgb(0.1, 0.9, 0.9),
+];
 const MIN_ENTITY_SIZE: Vec2 = Vec2::splat(0.02);
 const CAMERA_COLOR: Color = Color::rgb(0.9, 0.9, 0.9);
 
+#[derive(Resource, Debug)]
+struct PlayerColors {
```
Indy2222, line 31: I think that the code would be a bit nicer if this becomes `PlayerColors([Color])` and has methods `get_color(Player, ActiveObjectType)`. `get_color()` would then pull the base player color from the hash map and then use a simple `match` on the object type to derive the object color from the base player color.

The benefit of this approach is that:

* `draw_entities_system()` system would be simpler
* we could reuse the struct elsewhere easier in the future (at no cost of extra complexity)
* `player.to_num()` conversion would happen as late as possible (inside `get_color()` method)

crates/controller/src/hud/minimap/fill.rs
```diff
@@ -6,22 +9,56 @@ use de_core::{
...
+    player: (Color, AHashMap<ObjectType, Color>),
+    enemies: [(Color, AHashMap<ObjectType, Color>); 4],
```
Indy2222, line 33: Please replace the hard-coded 4 with `Player::MAX_PLAYERS` so that the code does not break when max players is changed.

Indy2222 (changes requested):
Thank you for the PR!

I have left some comments requesting some restructuring of the implementation but it should not be difficult.

crates/controller/src/hud/minimap/fill.rs
Indy2222: > Thanks for the guidance! Forgive my noobness. I'm not sure if you meant an array or a slice of `Color` by `PlayerColors([Color])`. I think for slices I would need to manage lifetimes and I'm not confident on how to do that well yet, so I stuck with an array. If it really needs to be a slice I can do more learning and give that a go.

No need for forgiveness :heart: 

I meant array.

> I also couldn't figure out a way to implement a tuple struct where `get_color()` wouldn't be deriving the unit color on every invocation, and when I do that the game slows down tremendously.

It is a bit surprising that it is slow. You basically just convert player to number, take the base color from the array and then "multiply" RGB values in a match.

> So I refactored the closest to what I think you wanted, though the current player is no longer always blue, instead each player gets the color of their player number. It does see more re-usable now, but with `PlayerColors::new([Color; size])` instead of `PlayerColors([Color; size])`.

Ok. I will do another review.

> If you want the playable player to always be blue still, I can refactor again easily and either add a separate method for `get_playable_color()`, or pass `game` or a playable boolean into the existing `get_color()`

It is not necessary. Just make sure that we do not mention the blue in any documentation.

> If there is a way to have tuple struct like `PlayerColors([Color])` have some internal persistent state to cache the derived colors that I just don't know how to do, let me know or point me to some docs.

crates/controller/src/hud/minimap/fill.rs
```diff
@@ -1,27 +1,75 @@
...
+                derive_colors(colors[2]),
+                derive_colors(colors[3]),
+            ],
+        }
+    }
+
+    fn get_color(&self, player: Player, object_type: ObjectType) -> Color {
+        let player_idx: usize = (player.to_num() - 1) as usize;
+        let default_color: Color = self.player_colors[player_idx];
+        let object_color = self.object_colors[player_idx].get(&object_type);
```
Indy2222, line 50: Instead of having a hash map, I would create a `match` here and multiply rgb values of default color with different factors in each match arm. That would be way faster. Example for inspiration:

```rust
let player_idx: usize = (player.to_num() - 1) as usize;
let player_color = self.player_colors[player_idx].as_hsla();
match object_type {
    ActiveObjectType::Building => player_color,
    ActiveObjectType::Unit(_) => player_color.with_s(0.7 * player_color.s())
}
```

crates/controller/src/hud/minimap/fill.rs
```diff
@@ -1,27 +1,75 @@
...
+                derive_colors(colors[1]),
+                derive_colors(colors[2]),
+                derive_colors(colors[3]),
+            ],
+        }
+    }
+
+    fn get_color(&self, player: Player, object_type: ObjectType) -> Color {
+        let player_idx: usize = (player.to_num() - 1) as usize;
+        let default_color: Color = self.player_colors[player_idx];
```
Indy2222, line 49: I would change variable name from `default_color` to `player_color`.

crates/controller/src/hud/minimap/fill.rs
```diff
@@ -1,27 +1,75 @@
...
+            object_colors: [
+                derive_colors(colors[0]),
+                derive_colors(colors[1]),
+                derive_colors(colors[2]),
+                derive_colors(colors[3]),
+            ],
+        }
+    }
+
+    fn get_color(&self, player: Player, object_type: ObjectType) -> Color {
```
Indy2222, line 47: Please change type of `object_type` to `ActiveObjectType` (other object types are not colored) with this.

Indy2222 (changes requested):

crates/controller/src/hud/minimap/fill.rs
redbugz: I tried the tuple struct with your simpler algorithm and timed the `get_color()` call, and it's actually faster than doing the hash lookup, which surprised me. I'm not sure why mine was so slow before, perhaps because I was doing a clone on the color?
So it is back to the simplified tuple struct and is not slow.

I wasn't sure how to get the ObjectType coerced to an ActiveObjectType before, but I found a way that works, let me know if there is a more idiomatic way to do it than what I did.

crates/controller/src/hud/minimap/fill.rs
redbugz: this implementation was much faster than mine and made the tuple struct work

crates/controller/src/hud/minimap/fill.rs
Indy2222: Computing the hash and accessing the hash table is actually pretty slow, even though AHashMap is one of the fast ones.

Indy2222 (commented):
One minor formatting note. Otherwise, LGTM. I'll test it later.

Indy2222:
Also, please note that I have done some important CI related fixes to `main`. I suggest that you merge `main` into this branch (if you have not done yet already).

Indy2222 (approved):

